PR #19, opened 2016-03-16
Fix for clearing a MediaQueue containing a disposed Song.

julianhobler:
Steps to reproduce the issue:
- Play Song A using MediaPlayer
- Call MediaPlayer.Stop() and dispose Song A
- Notice that MediaPlayer's MediaQueue still holds a reference to Song A
- Try to play Song B using MediaPlayer

MediaPlayer.Play() clears the MediaQueue which executes the Stop() method for all its Songs, which could be disposed already.

### DIFF
--- a/src/Media/Xiph/Song.cs
+++ b/src/Media/Xiph/Song.cs
@@ -286,12 +286,12 @@ namespace Microsoft.Xna.Framework.Media
 			timer.Stop();
 			timer.Reset();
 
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Fix the tabbing and you're all set.

julianhobler:
There we go.

flibitijibibo:
Beautiful, thanks!

Issue #24, opened 2016-05-06
Mojoshader missing variable, and assertion in preShader


jeffphilp:
Hi,

When initially loading the following file:
[shader.txt](https://github.com/FNA-XNA/FNA/files/252239/shader.txt)

it fails to compile the generated glsl because of a missing variable ps_v1, I believe this is caused by the NORMAL0 parameter passed into pixel shader which I think can be fixed by adding the following code:
     
     else if (usage == MOJOSHADER_USAGE_NORMAL)
     {
         push_output(ctx, &ctx->globals);
         output_line(ctx, "vec4 %s;", var);
         pop_output(ctx);
     } // else if        

at https://github.com/flibitijibibo/MojoShader/blob/master/mojoshader.c#L2731

This change allowed the shader to compile, but when I actually make use of the effect I run into another issue which relates to an assertion failure in the MOJOSHADER_runPreshader
https://github.com/flibitijibibo/MojoShader/blob/master/mojoshader_effects.c#L193
where the final operand is actually of type MOJOSHADER_PRESHADEROPERAND_INPUT.  
The last commit to that line is putting in the assertion for OUTPUT only though the previous changeset had a commented out assert which tolerated INPUT or OUTPUT.  Should we be allowing INPUT types as the destination?

flibitijibibo:
I think I fixed the first one (it was the same realm as the COLOR2 issue), but the preshader may need some different attention.

As far as I know, the VM does not write to input registers, and the preshader instructions indicate this:

```
            neg r0.x, c1.x
            cmp c0.x, r0.x, (0), (1)
            neg r0.xyz, c0.xyz
            add c1.xyz, (1, 1, 1), r0.xyz
```
r0 is a temp register and c0/c1 are output registers. See if you can figure out why it wants to write to input... (or optimize this out of the preshader, if you want).

jeffphilp:
Forgive my ignorance, I am fairly new to the graphics area, but by VM do you mean a layer provided by mojoshader or by the effects framework?  Also how were you able to generate those instructions?

flibitijibibo:
The preshader VM is this guy:

https://github.com/flibitijibibo/MojoShader/blob/master/mojoshader_effects.c#L17

[...]

PR #25, opened 2016-05-06
Log any errors from mojoshader after calling MOJOSHADER_parseEffect.

jeffphilp:
After discussion in https://github.com/FNA-XNA/FNA/issues/24 this pull request checks the error state after calling parseEffect and logs it out.


flibitijibibo:
Cool, thanks for looking into this. I merged this manually with the following changes:

- Changed Win32 line endings to Unix line endings
- Loop to go through all the errors stored in the effect
- Use pointers instead of PtrToStructure

Commit: https://github.com/FNA-XNA/FNA/commit/45abca0735ad87f904aae6ee6ebbc4045e6bb360

Issue #39, opened 2016-07-14
Unexpected result from Vector2.Transform(ref Vector2 value, ref Quaternion rotation, out Vector2 result)


EnoughTea:
Hello!

It seems that method `Vector2.Transform(ref Vector2 value, ref Quaternion rotation, out Vector2 result)` returns strange results when pitch and roll is involved. In MonoGame it works as expected.

Demo:

```
var vec2 = new Vector2(10, 20); 								
var quaternion = Quaternion.CreateFromYawPitchRoll(3.1415f, 1.5707f, 1.5707f);
var matrix = Matrix.CreateFromYawPitchRoll(3.1415f, 1.5707f, 1.5707f);

Vector2 rotatedByQ = Vector2.Transform(vec2, quaternion);
Vector2 rotatedByM = Vector2.Transform(vec2, matrix);

// Next line will throw; rotation by matrix will work as expected, 
// while rotation by quaternion will be different:
Assert.That(rotatedByQ, Is.EqualTo(rotatedByM));
```

_quaternion_'s matrix is identical to _matrix_, so `Vector2.Transform(vec2, Matrix.CreateFromQuaternion(quaternion))` works as expected. This leaves the only culprit: `Vector2.Transform(Vector2, Quaternion)` method.

Turning Vector2 into Vector3 and using its `Transform` works; replacing `Vector2.Transform` code with MonoGame's one works as well.

I took a peek inside `Vector2.Transform`, it uses a canonical formula _v' = q * v * q^-1_, and all quaternion methods look good at least from first glance, so I'm kinda stumped.

flibitijibibo:
What happens when you adapt Vector3.Transform to Vector2? Take this function and rewrite Vector2's:

https://github.com/FNA-XNA/FNA/blob/master/src/Vector3.cs#L1149

If adapting this works then go ahead and send out a PR, and I'll merge it.

EnoughTea:
Yep, it works.

flibitijibibo:
Solved by #40, thanks again for checking it out.

PR #40, opened 2016-07-14
Vector2.Transform(Quaternion) fix.

EnoughTea:
Vector2.Transform now works with quaternions as expected, related issue:
https://github.com/FNA-XNA/FNA/issues/39


flibitijibibo:
Beautiful, thanks for looking at this.

PR #41, opened 2016-07-14
Fixed invalid parameter order in Vector4.Hermite.

EnoughTea:
In the non-ref `Vector4.Hermite` overload there is a typo: W is going in first into the resulting vector's ctor instead of X.


flibitijibibo:
Not even going to ask how that happened...

PR #55, opened 2016-08-25
Normalized* family of packed vectors does not work for negative values.

EnoughTea:
Normalized* family of packed vectors (namely NormalizedByte2, NormalizedByte4, NormalizedShort2, NormalizedShort4) does not pack negative values correctly. It looks like Pack() method drops all bytes after the first one for values < 0.

XNA docs for Normalized* state that it is a "Packed vector type containing * 8-bit signed normalized values, ranging from −1 to 1.", so negative values should be supported.

Ways to reproduce:
Just convert any negative vectors between each other
```
Vector2 target = Vector2(-1, -1);
Assert.AreEqual(target , new NormalizedShort2(target).ToVector2());  // throws
```
Also by looking at the value: `-Vector4.One` packed into 4 bytes should look like 0x81818181, but instead I've got `0xFFFFFF81`

How to fix:
I am not very good at bit math, so I just grabbed Pack() code from Monogame, it seems to work for all test cases.

### DIFF
--- a/src/Graphics/PackedVector/NormalizedByte2.cs
+++ b/src/Graphics/PackedVector/NormalizedByte2.cs
@@ -57,8 +57,8 @@ namespace Microsoft.Xna.Framework.Graphics.PackedVector
 		public Vector2 ToVector2()
 		{
 			return new Vector2(
-				((sbyte) (packedValue & 0xFF)) / 127.0f,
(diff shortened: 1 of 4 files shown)


flibitijibibo:
Merged with an additional style fix, those packing lines are getting crazy. Thanks for looking at this!

PR #64, opened 2016-10-22
Reduced some attachment array garbage generation in SetRenderTargets

PumpkinPaul:


PumpkinPaul:
I did think about just creating the array up front. 

I could create a honey but there's this in the constructor...

currentAttachments = new uint[numAttachments];

Should we just make attachments the same size?

PumpkinPaul:
Honey? That should be *const* - haha

flibitijibibo:
Yeah, let's alloc that too. To prevent false positives we can just set all unused target slots to uint.MaxValue, like we do when deleting attached targets in the GLDevice.

flibitijibibo:
Actually, odds are good that max draw buffers is way higher than 4, so we can keep the code basically as-is. Extra credit would be checking if it's _less_ than 4, which is unlikely but possible for super old hardware.

flibitijibibo:
(T-t-triple post, for reference:)

https://people.freedesktop.org/~imirkin/glxinfo/

flibitijibibo:
One more reference sheet: https://developer.apple.com/opengl/capabilities/GLInfo_1075.html

So basically the only card that does NOT have at least 4 draw buffers is the GMA 950 on 10.7 Lion, a card that got cut out for Mountain Lion, and all other cards have at least 8 on 10.8 and higher. The only card that doesn't have 8 in Mesa's world is an Adreno GPU and the NVIDIA GeForce FX, a card so old that it has Windows 98 drivers. I suppose it couldn't hurt to do Math.Min(GraphicsDevice.MAX_RENDERTARGETS, maxDrawBuffers) like we do with sampler objects, but the main point is that we can arbitrarily target 4 even in the GLDevice and be safe almost 100% of the time. If we ever hit a <4 case, it's almost certainly something crazy like a Radeon Xpress 200M.

flibitijibibo:
Here's my version of the garbage reduction applied to ModernGLDevice, can you mash this up with your PR and check to make sure it still works? For OpenGLDevice we can still use MAX_DRAW_BUFFERS with min(), as mentioned earlier.

```
diff --git a/src/FNAPlatform/ModernGLDevice.cs b/src/FNAPlatform/ModernGLDevice.cs
index edd33d8..0ad3f6d 100644
--- a/src/FNAPlatform/ModernGLDevice.cs
+++ b/src/FNAPlatform/ModernGLDevice.cs
@@ -405,6 +405,8 @@ namespace Microsoft.Xna.Framework.Graphics
 		private GLenum[] drawBuffersArray;
 		private uint currentRenderbuffer;
 		private DepthFormat currentDepthStencilFormat;
+		private uint[] attachments = new uint[GraphicsDevice.MAX_RENDERTARGET_BINDINGS];
+		private GLenum[] attachmentTypes = new GLenum[GraphicsDevice.MAX_RENDERTARGET_BINDINGS];
 
 		#endregion
 
@@ -691,12 +693,10 @@ namespace Microsoft.Xna.Framework.Graphics
 			}
 
 			// Initialize render target FBO and state arrays
-			int numAttachments;
-			glGetIntegerv(GLenum.GL_MAX_DRAW_BUFFERS, out numAttachments);
-			currentAttachments = new uint[numAttachments];
-			currentAttachmentTypes = new GLenum[numAttachments];
-			drawBuffersArray = new GLenum[numAttachments];
-			for (int i = 0; i < numAttachments; i += 1)
+			currentAttachments = new uint[GraphicsDevice.MAX_RENDERTARGET_BINDINGS];
+			currentAttachmentTypes = new GLenum[currentAttachments.Length];
+			drawBuffersArray = new GLenum[currentAttachments.Length];
+			for (int i = 0; i < currentAttachments.Length; i += 1)
 			{
 				currentAttachments[i] = 0;
 				currentAttachmentTypes[i] = GLenum.GL_TEXTURE_2D;
@@ -3240,8 +3240,6 @@ namespace Microsoft.Xna.Framework.Graphics
 			}
 
 			int i;
-			uint[] attachments = new uint[renderTargets.Length];
-			GLenum[] attachmentTypes = new GLenum[renderTargets.Length];
 			for (i = 0; i < renderTargets.Length; i += 1)
 			{
 				IGLRenderbuffer colorBuffer = (renderTargets[i].RenderTarget as IRenderTarget).ColorBuffer;
@@ -3265,7 +3263,7 @@ namespace Microsoft.Xna.Framework.Graphics
 			}
 
 			// Update the color attachments, DrawBuffers state
-			for (i = 0; i < attachments.Length; i += 1)
+			for (i = 0; i < renderTargets.Length; i += 1)
 			{
 				if (attachments[i] != currentAttachments[i])
 				{
@@ -3355,14 +3353,14 @@ namespace Microsoft.Xna.Framework.Graphics
 				}
 				i += 1;
 			}
-			if (attachments.Length != currentDrawBuffers)
+			if (renderTargets.Length != currentDrawBuffers)
 			{
 				glNamedFramebufferDrawBuffers(
 					targetFramebuffer,
-					attachments.Length,
+					renderTargets.Length,
 					drawBuffersArray
 				);
-				currentDrawBuffers = attachments.Length;
+				currentDrawBuffers = renderTargets.Length;
 			}
 
 			// Update the depth/stencil attachment
diff --git a/src/Graphics/GraphicsDevice.cs b/src/Graphics/GraphicsDevice.cs
index 641ab93..dc5fef8 100644
--- a/src/Graphics/GraphicsDevice.cs
+++ b/src/Graphics/GraphicsDevice.cs
@@ -308,8 +308,9 @@ namespace Microsoft.Xna.Framework.Graphics
 
 		#region Private RenderTarget Variables
 
-		// 4, per XNA4 HiDef spec
-		private readonly RenderTargetBinding[] renderTargetBindings = new RenderTargetBinding[4];
+		// Per XNA4 HiDef spec
+		internal const int MAX_RENDERTARGET_BINDINGS = 4;
+		private readonly RenderTargetBinding[] renderTargetBindings = new RenderTargetBinding[MAX_RENDERTARGET_BINDINGS];
 
 		// Used to prevent allocs on SetRenderTarget()
 		private readonly RenderTargetBinding[] singleTargetCache = new RenderTargetBinding[1];
```

PumpkinPaul:
Mashed your changes into the PR and clamped the draw buffers to a maximum of 4

flibitijibibo:
Merged this manually along with some similar work for vertex attributes and sampler objects:

https://github.com/FNA-XNA/FNA/commit/3e5cdb8ae57b772e47190d0028727c3f97d0e3c2

PR #67, opened 2016-11-01
Vector exact float equality

### LINKED ISSUE #66: Vector2 == should not use epsilon check
I know from experience that the best way to get your bugs solved is to be positive, so right off the bat, I love FNA!

Now to the point, XNA does not use epsilon checks when comparing Vectors (Equals and operator==), it uses the built in floating point equality which has a well defined behavior.

I know the reasoning behind floating point epsilons, however generally they should only be applied on a case by case basis, rather than a blanket over all floating point comparisons.

The current implementation causes unexpected behaviors, such as comparing a Vector2 initialized with Single.PositiveInfinity to compare incorrectly with itself (infinity comparison is well defined). Another example is if you were to use the Vector2 as a key in a Dictionary, the GetHashCode becomes inconsistent with the equality tests.

TrevorSundberg:
Fixed #66

### DIFF
--- a/src/Vector2.cs
+++ b/src/Vector2.cs
@@ -160,8 +160,9 @@ namespace Microsoft.Xna.Framework
 		/// <returns><c>true</c> if the instances are equal; <c>false</c> otherwise.</returns>
 		public bool Equals(Vector2 other)
 		{
-			return (	(MathHelper.WithinEpsilon(X, other.X)) &&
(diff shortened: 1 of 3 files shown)


flibitijibibo:
Excellent, thanks for the speediness. For extra credit you can verify this with the other surface-level uses of WithinEpsilon, but for now...

PR #75, opened 2016-11-24
Pause resume subcategories

kiates:
I verified this behavior by comparing an XNA/XACT build against a FNA/FACT build on Windows 10.


flibitijibibo:
Beautiful, thanks! Looking forward to your other XACT fixes because hoo boy there are a lot of em

Issue #76, opened 2016-12-03
Various pitch related features are not working


kiates:
It appears that the conversion from encoded pitch values (individual and ranges) is off by 10.  Encoded 12 semitones is 1200, however on reading these pitch values are divided by 1000, thus 1200 results in 1.2 instead of 12.

Additionally the OpenAL audio framework seems to need a scaled -1 to 1 range for +/- one octave, while XACT works with -12 to 12. Applying this conversion within the OpenAL wrapper corrects the mismatch.

One final observation is when calculating pitch variations at the Play wave level, the sound level should always be added regardless of whether add/replace was occurring at the play wave level.

kiates:
I have a branch with the above corrections and will submit a pull request for your consideration.

flibitijibibo:
https://github.com/FNA-XNA/FNA/commit/c6e964e4e2aa5d6197fafc7daa34d2debcb70ce8

PR #77, opened 2016-12-03
Fix various discovered pitch issues

kiates:
This pull request addresses the issued summarized in #76.

Using a trivial test application I compared and contrasted the XACT engine against the FACT (is that the right name?) for the aspects of pitch covered.

### DIFF
--- a/src/Audio/Cue.cs
+++ b/src/Audio/Cue.cs
@@ -735,12 +735,16 @@ namespace Microsoft.Xna.Framework.Audio
 				/* The final pitch should be the combination of the
 				 * authored pitch and RPC/Event pitch results.
 				 */
-				INTERNAL_instancePool[i].Pitch = (
+
+				float aggregatePitch = (
 					INTERNAL_instancePitches[i] +
 					rpcPitch +
 					eventPitch +
 					INTERNAL_rpcTrackPitches[i]
 				);
+
+				// XACT uses -12 to 12 (semitones) internally for pitch, OpenAL uses -1 to 1 (octaves).
+				INTERNAL_instancePool[i].Pitch = aggregatePitch/12.0f;
 
 				/* The final filter is determined by the instance's filter type,
 				 * in addition to our calculation of the HF/LF gain values.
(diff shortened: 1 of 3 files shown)


flibitijibibo:
Seems Pitch gets calculated differently from SoundEffect's way, much like Volume. I only ever looked at the xap raw text, so it looked like it was -1000/1000 to look like SoundEffectInstance.Pitch's -1/1. So we get to fix the division, but we need to be sure it's done consistently throughout the engine and doesn't get done multiple times to the same value by accident. Other than that, can confirm the -1200/1200 storage range in the xap.

kiates:
Before submitting this, I did a search for "/ 1000" type patterns to see if there were any other places related to pitch.  I didn't find any.  I'll double check that aspect.

Agreed that it needs to be consistent and there are a few more places where pitches are aggregated before setting the SoundEffectInstance.Pitch setter.  I'll move the converstion from -12/12 to -1/1 to that setter and try and hit the other points to verify the input ranges.

kiates:
Take a look and see if this is a cleaner approach.  I created a test case that has a initial non-zero pitch, has a pitch variation that adds 1 semitone per loop, and has a single pitch event that at a certain time adds another semitone.  This matches XNA exactly in my audio test.

flibitijibibo:
It's cleaner, but breaks accuracy for SoundEffectInstance outside of XACT. We have to do the division on the XACT side, other than that it all LGTM.

kiates:
Can I use INTERNAL_isXACTSource inside SoundEffectInstance to guard the conversion?

kiates:
`					// XACT uses -12 to 12 (semitones) internally for pitch, OpenAL uses -1 to 1 (octaves).
					float finalPitch = INTERNAL_isXACTSource ? value / 12.0f : value;

					AudioDevice.ALDevice.SetSourcePitch(
						INTERNAL_alSource,
						finalPitch,
						!INTERNAL_isXACTSource
					);`

flibitijibibo:
I'd like to avoid using isXACTSource as much as possible; it's kind of a hack that really shouldn't exist to begin with. Plus, we want as much XACT stuff to be in the actual XACT code as possible, particularly for when it all gets converted to native code later on (and for whenever someone's researching XACT and doesn't care about the XNA part of it).

kiates:
And when you refactor to remove the hacks, you could take this out as well. ;) At least it would all be together.

It seems from my cursory understanding of the code, that you need a generic wrapper for a sound effect that SoundEffectInstance and a future XactSoundEffectInstance can use, since they seem to have differing behavior on several points (Clamping, Volume, Pitch, Apply3D, etc.).  Some of those could be solved with a simple XACT struct wrapper around SoundEffectInstance to adjust inputs/outputs, but that won't solve all of them. The generic wrapper would need to expose configurable policies for some of these basic things which is one step up from a isXACT flag.

It looks like Xact even goes down into the IALSource.GenSource to force max gain.

I'm not in position to take this kind of tangent, but I can move the division to the call sites, but it's even uglier IMHO.

flibitijibibo:
My main goal for FACT will be to have no intermediate layer between XACT and the sound backend, and the SoundEffect API will be entirely separate structures in the library as an extra feature. It's definitely true that the SFI and XACT subsystems are mixed pretty poorly, since SFI has way more limitations and we're basically stripping all that out to avoid having to write a second class.

For now, the main reason to avoid further hacks in opposite subsystems is to keep them decoupled for when the rewrite happens. The more they're coupled, the harder it will be to unravel it all when it comes time to write the FACT mixer without effectively rewriting the existing code first.

Either one works, but if it ever bothers me then the first thing I'll end up doing is replacing it and risk introducing new bugs (and since they're coupled, they could be in either system rather than just one of them, or worse, both!).

kiates:
Here's a version with the call sites taking the responsibility.

PR #79, opened 2016-12-03
Doppler

### LINKED ISSUE #78: DopplerPitchScalar and OrientationAngle variables not implemented
These two XACT internally computed variables are necessary for various effects using the RPC functionality.

kiates:
This resolves #78 

Using a trivial test application I compared and contrasted XACT vs my implementation in FACT.  The test only moved the emitter souce along one dimension (across the X-axis some distance in front of the listener, but the mathematics should hold for all 3 dimensions).  I was able to compare the values in both variables for accuracy, as well as listen to the effect between the two implementations.  It seems an exact match.

### DIFF
--- a/src/Audio/Cue.cs
+++ b/src/Audio/Cue.cs
@@ -296,8 +296,51 @@ namespace Microsoft.Xna.Framework.Audio
 					INTERNAL_listener.Position
 				)
 			);
-			// TODO: DopplerPitchScaler, OrientationAngle
+
+			// Doppler
+			SetVariable(
+				"DopplerPitchScalar",
+				 CalculateDoppler()
+			);
+
+			// TODO: OrientationAngle
 			INTERNAL_isPositional = true;
+		}
+
+		private float CalculateDoppler()
+		{
+			// Adapted from algorithm published as a part of the webaudio specification:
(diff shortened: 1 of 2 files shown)


flibitijibibo:
This one looks good, it just has the pitch commits behind it.

flibitijibibo:
Merged manually with nitpicky style fixes:

https://github.com/FNA-XNA/FNA/commit/f10fb02e72cae98ff737fae345f3a9c452e161b8

flibitijibibo:
Digging this really old thing up... I'm looking into X3DAudio and found this from the [EMITTER docs](https://msdn.microsoft.com/en-us/library/windows/desktop/microsoft.directx_sdk.x3daudio.x3daudio_emitter(v=vs.85).aspx):

>Orientation of the front direction. This value must be orthonormal with OrientTop. OrientFront must be normalized when used. For single-channel emitters without cones OrientFront is only **used for emitter angle calculations**.

It would seem that our `OrientationAngle` math isn't entirely complete; in addition to doing some math for Listener.Forward we need to do some for Emitter.Forward as well. @kiates, would it be easy enough to add this to FNA's existing math? I can transfer it over to FAudio, but I don't know exactly how to adjust this.

Issue #80, opened 2016-12-03
[XACT] FNA seems significantly quieter than XNA


kiates:
Adjusting the volume by a factor of 2 in the OpenAL wrapper seems to normalize the levels.  I don't know why, but 2x equals 6db, so there may be a 6db error somewhere.  I measured the difference between XACT FACT running the same test application using a white noise sample with a simple iphone based sound pressure meter and they were extremely close after the 2x adjustment.

I could only verify this on windows at the moment, so I'm not sure if there could be platform differences.

kiates:
I'm still trying to find root cause for this.  What I have learned so far is there is a difference in audio levels between positional and non-positional sounds in XACT (non-positional are louder), but in FNA is there is no difference.  In additon, positional sounds in XACT and FNA are approximately equal in my comparisons, so the issue is that XACT non-positional sounds are inherently louder for some reason.

FNA simulates non-positional sounds by staying within the clamped area of the default OpenAL automatic distance attenuation model (Inverse Clamped).  This appears to be working fine, as I temporarily turned of the distance model to compare and the volume levels remained the same.

Further analysis surrounding why XACT non-positional sounds are louder is necessary.

flibitijibibo:
My guess is that this is part of XAudio2 behavior. I really wish I could trace XAudio2 calls, as that would tell us everything we need to know.

flibitijibibo:
Likely relevant to this is the IXACT3Cue interface:

https://github.com/hrydgard/minidx9/blob/master/Include/xact3.h#L939

Apply3D probably calls a bunch of things that are otherwise uncalled, meaning Cues are treated differently without the added mix information.

flibitijibibo:
The latest commits do a lot of scary things to volume storage and calculation, so let's make sure we're at least as "good" as we were before (or possibly better if wizardry occurred).

kiates:
Ok, I'll integrate your new changes and compare/contrast.

flibitijibibo:
Closing as WONTFIX in the existing audio implementation, and WAITING for the FAudio rewrite coming up.

I'm pretty sure this was caused by the sound having different volumes with the default matrix coefficients, followed by X3DAudioCalculate doing a bunch of things alongside whatever XACT volume RPCs you might have. Once this pull request is merged, this should be done unless I'm wrong (test this whenever you can):

https://github.com/flibitijibibo/FACT/pull/8

PR #81, opened 2016-12-03
Volume differences

kiates:
This is in regard to issue #80 

I think OpenAL is used for all three platforms (Win, Linux, OSX), so I would expect the behavior to be consistent, and therefore my change working the same on all three.  I'm not 100% sure though.

### DIFF
--- a/src/Audio/Cue.cs
+++ b/src/Audio/Cue.cs
@@ -296,8 +296,51 @@ namespace Microsoft.Xna.Framework.Audio
 					INTERNAL_listener.Position
 				)
 			);
-			// TODO: DopplerPitchScaler, OrientationAngle
+
+			// Doppler
+			SetVariable(
+				"DopplerPitchScalar",
+				 CalculateDoppler()
+			);
+
+			// TODO: OrientationAngle
 			INTERNAL_isPositional = true;
+		}
+
+		private float CalculateDoppler()
+		{
+			// Adapted from algorithm published as a part of the webaudio specification:
+			//   https://dvcs.w3.org/hg/audio/raw-file/tip/webaudio/specification.html#Spatialization-doppler-shift
+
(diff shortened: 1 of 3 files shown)


flibitijibibo:
This one definitely seemed odd, and sure enough, my ears got definitely blasted by a couple XACT games I had to test. This _might_ be Win32-specific; try the stock OpenAL Soft binary first...

http://kcat.strangesoft.net/openal-binaries/openal-soft-1.17.2-bin.zip

... and if that's still too soft, you may want to look into the Win32 backends:

https://github.com/kcat/openal-soft/tree/master/Alc/backends

If you want some more XACT games to test against, I have these games on Steam with 'fna-unstable' branches for Windows (sometimes the branch is named 'fna-win32'):

- Rogue Legacy
- Dust: An Elysian Tail
- Salt and Sanctuary

kiates:
I hadn't tried this fix in Solaroids yet, so I did a little XNA/FNA comparison on the same Win32 box.  It seems that the audio that doesn't use RPC presets (from what I can tell) plays louder with the fix (non-positional audio effects, and background music), but those that do use them play at the same level.  If this is the case it suggests the correction may be somewhere else other than at the OpenAL interface.

My testbench was focused on RPC aspects and didn't have any "simple" sounds.  I'll look into it more and update this pull request.

flibitijibibo:
The way I usually debug volunes is by going to the sfi.Volume set at the end of INTERNAL_update and just printf spam all the volume attribs that get multiplied into the final volume. Crude, but it tells you everything you need without a lot of stepping!

flibitijibibo:
Since the current patchset doesn't fix the accuracy issue and it has those other commits from previous PRs I'll go ahead and close this one, but we still have #80 open to dig into this further.

PR #85, opened 2016-12-11
Address various XACT cue state inaccuracies

kiates:


kiates:
My hunch is some of the boolean logic could be optimized, but it behaves correctly now using my test application built with XACT and then FNA.

flibitijibibo:
I may play with this a little first; I found some weird behaviors when debugging Owlboy so I need to merge this with what I ended up finding (like that IsStopped thing I mentioned a while back).

flibitijibibo:
Just pushed this fix for Owlboy, can you pull from mainline and see if your tests still work?

https://github.com/FNA-XNA/FNA/commit/2d60c48ba78a85d5957efcc7bf594c292fc8abeb

If everything still works I'll merge this in.

kiates:
I merged your change into my integration branch (my tests rely on a lot more than state changes to be fixed in order to run).  At least for my tests it seems to mimic the same behavior without the long boolean expression in the IsStopped property.

However, I improved my test to look scrutinize the IsStopped flag during the dispose process, and right after I call dispose, XACT reports IsStopped == false, while FACT reports IsStopped == true.  Is there a reason that your forcing IsStopped to true in this scenario? Mind you this scenario is pretty pathological, but while the patient is open so to speak, lets get it right. :)

BTW, my branch without your changes also has the same issue.

flibitijibibo:
No reason at all! In fact it only now occurs to me that Dispose probably free()s the IXACTCue entirely, so I imagine all the properties other than IsDisposed return false since the XACT call likely returns 0 when the IXACTCue* == NULL.

I just fixed that and merged this patch manually.

PR #91, opened 2016-12-14
Event types

kiates:
This body of work represents my effort to help support at least the parsing of all event types allowed on tracks attached to sounds. I wanted to get this pull request started to get feedback.

At this point it parses all event types I could create with the XACT3 tool (Feb 2010 build).  There are a couple of event types still in the code that are not handled, but I don't know what they are or how to create them.

For many of them I was also able to implement the run-time behavior for the events:

- Stop - Immediate/AsAuthored and Cue scope.  Track scope is not implemented.
- Pitch/Volume - SettingType of Equation, Equation Types of Value and Random, Operations of Replace and Add. Ramp is parsed but not implemented. Repeat behavior is parsed but not implemented.
- Marker - Parsed but not implemented.

Alas, there is still more work to be done to support run-time behavior for the following that was also mentioned above:

- Repeating Events - They are parsed and the data is available, but the mechanics to reshedule the events has not been started.
- Ramp Setting Type - These are parsed and the data is available, but the bounding of the event between the timestamp and the duration, and the respective interpolation are not implemented yet.
- Markers - I've never used these myself, and my understanding is they don't necessarily work correctly in XACT either.

The code could be refactored to avoid duplication as a next step.  But it's a start, and what has been implemented passes my comparison tests. It looks like Play wave also suffers from this.

### DIFF
--- a/src/Audio/CueData.cs
+++ b/src/Audio/CueData.cs
@@ -853,7 +853,7 @@ namespace Microsoft.Xna.Framework.Audio
 		{
 			foreach (XACTEvent curEvent in Events)
 			{
-				if (curEvent.Type == 1)
+				if (curEvent.Type == XACTEvent.EventTypeCode.PlayWave)
 				{
 					((PlayWaveEvent) curEvent).LoadTracks(
 						audioEngine,
@@ -866,10 +866,17 @@ namespace Microsoft.Xna.Framework.Audio
 
 	internal abstract class XACTEvent
 	{
-		public uint Type
+		public enum EventTypeCode
+		{
+			PlayWave = 1,
+			SetVolume = 2,
+			SetEquationPitch = 3,
+			SetRandomPitch = 4
+		}
+
+		public EventTypeCode Type
 		{
 			get;
-			private set;
 		}
 
 		public uint Timestamp
@@ -884,12 +891,12 @@ namespace Microsoft.Xna.Framework.Audio
 
 		protected static readonly Random random = new Random();
 
-		public XACTEvent(uint type, uint timestamp)
+		public XACTEvent(EventTypeCode type, uint timestamp)
 			: this(type, timestamp, 0, 0)
 		{
 		}
 
-		protected XACTEvent(uint type, uint timestamp, int count, float frequency)			
+		protected XACTEvent(EventTypeCode type, uint timestamp, int count, float frequency)			
 		{
 			Type = type;
 			Timestamp = timestamp;
@@ -950,7 +957,7 @@ namespace Microsoft.Xna.Framework.Audio
 			ushort trackVariationType,
 			bool trackVariationOnLoop,
 			byte[] weights
-		) : base(1, timestamp) {
+		) : base(EventTypeCode.PlayWave, timestamp) {
 			INTERNAL_tracks = tracks;
 			INTERNAL_waveBanks = waveBanks;
 			INTERNAL_minPitch = minPitch;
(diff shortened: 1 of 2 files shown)


src/Audio/CueData.cs
```diff
@@ -663,36 +682,98 @@ public XACTClip(BinaryReader reader, double clipVolume, byte filterType)
 						weights
 					);
 				}
-				else if (eventType == 7) // SetPitchEvent
```
kiates: Note that I converted these eventType constants to hex in the hopes that I would see  a bit pattern relating the non-repeating with the repeating codes, but there doesn't seem to be much of a relationship, just arbitrary event codes.  I think rather than change them back, I would just move forward to an enum to represent all known codes.

src/Audio/CueData.cs
```diff
@@ -789,12 +878,23 @@ public uint Timestamp
 			private set;
 		}
 
-		protected static Random random = new Random();
+		public int Count { get; private set; }
```
kiates: Now with count and frequency available, we could update the timestamp after an event fires to schedule it again.

src/Audio/CueData.cs
```diff
@@ -740,10 +821,12 @@ public XACTClip(BinaryReader reader, double clipVolume, byte filterType)
 				else if (eventType == 15) // ???
```
kiates: Do you know how to create an eventType of 15 in XACT?

src/Audio/CueData.cs
```diff
@@ -740,10 +821,12 @@ public XACTClip(BinaryReader reader, double clipVolume, byte filterType)
...
 				{
 					// TODO: Codename OhGodNo -flibit
+					Events[i] = new NullEvent(eventTimestamp);
 				}
 				else if (eventType == 17) // Volume Repeat Event
```
kiates: Do you also know how to create an eventType of 17?

flibitijibibo: Looks like you already got 0x11, so no worries here...

src/Audio/CueData.cs
flibitijibibo: We may have to write it somewhere else since this data is technically supposed to be read-only, but the idea definitely works!

src/Audio/CueData.cs
flibitijibibo: Yeah, let's do the enum. At this point I'm 99% sure it's just a long-ass enum that feature-creeped as XACT got further into development. The recurrence enums are probably just enums copypasta'd after the pre-existing enums to add the recurrence feature.

src/Audio/CueData.cs
kiates: Lol. Right!

src/Audio/CueData.cs
kiates: And I thought that OhGodNo was just you being funny.

src/Audio/CueData.cs
```diff
@@ -777,24 +962,66 @@ public void LoadTracks(AudioEngine audioEngine, List<string> waveBankNames)
...
 			get;
 			private set;
 		}
 
-		public uint Timestamp
+		public int Count { get; private set; }
+
+		public float Frequency { get; private set; }
+
+		public Cue Cue { get; set; }
```
kiates: Should be able to.

flibitijibibo: Cool, let's move this Cue instance over to a local variable in Apply() and add the aforementioned enum, then I'll go ahead and merge this.

flibitijibibo:
Merged with a cleanup commit on top:

https://github.com/FNA-XNA/FNA/commit/9617671071a74bee746794c70f01b9a07337703b

It _should_ be functionally identical, but it's a pretty big cleanup so it may have one or two mistakes inside. Normally I wouldn't do this much on top, but I'm probably going to be going line-by-line for FACT when that happens so I compressed the line widths a whole bunch.

PR #92, opened 2016-12-16
Fix additive volume calculations. They should be in decibel space.

kiates:
After I got all the event types plumbed in I noticed that the volume calculations did not match XACT when add mode was used. I noticed this on the looping with variation.  This fixes that.  The event version most likely suffers from the same issue, so I'll be pushing an additional commit soon.

### DIFF
--- a/src/Audio/CueData.cs
+++ b/src/Audio/CueData.cs
@@ -875,7 +875,7 @@ namespace Microsoft.Xna.Framework.Audio
 						if (equationType == XactEventEquationType.Value)
 						{
 							// Absolute or relative value to set to.
-							float eventValue = XACTCalculator.CalculateAmplitudeRatio(reader.ReadSingle() / 100.0f);
+							float eventValue = reader.ReadSingle() / 100.0f;
 
 							// Unused/unknown trailing bytes.
 							reader.ReadBytes(9);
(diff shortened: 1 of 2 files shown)


flibitijibibo:
Because prevVolume comes from INTERNAL_instanceVolumes, which is just the float from GenerateInstance, I wonder if we should just store the dB value and send that value to PlayWave to avoid precision loss...? I'm kinda paranoid about volume precision, since we're already pretty far off and the XACT tools absolutely _love_ dropping volume precision for some reason.

kiates:
Yes changing all of FACT to store db, and then conversion once at the the AL interface would be best once all the wrinkles are worked out.

kiates:
I can look at making that change in a future commit once all the pieces are in place.  I still have Attack/Release to finish for a pull request.

flibitijibibo:
I'll focus on this one tomorrow as I have some data for it, go ahead and focus on the Attack/Release stuff with what you've got and we can work the new volume modifiers into the final version once we know how it fits in with the existing ones. If you want to wait until I have this fixed, you can send stuff from the other branches in the meantime (like the Variable access flag stuff).

flibitijibibo:
This has been replaced by a commit that stores everything as dB until the final Volume calculation:

https://github.com/FNA-XNA/FNA/commit/c1ad52c124bf928fea54f2de37216d649e315e24

... Is it just me or is this MUCH cleaner than before? o_o

kiates:
Oh yeah! Much simpler.

PR #94, opened 2016-12-26
Event system and RPC enhancements

kiates:
Happy Holidays FNA!

Refactor event system:
- Separate event instances from constant event parameters
- Movement towards unified parsing, avoiding duplication.
- Operation of volume at Db and Pitch in semitones. Conversion from fixed point units used for encoding handled at parse time and at SFI boundary.

Event enhancements:
- Support ramp event type.  All ramps are currently interpreted as linear for now (inclusing of curvature can easily be added at any time.).
- Support repeating events for volume/pitch.

RPC enhancements:
- Support Attack / Release RPC curves for track targets.

### DIFF
--- a/src/Audio/Cue.cs
+++ b/src/Audio/Cue.cs
@@ -75,7 +75,8 @@ namespace Microsoft.Xna.Framework.Audio
 		{
 			get
 			{
-				return INTERNAL_fadeMode == FadeMode.FadeOut;
+				return INTERNAL_fadeMode == FadeMode.FadeOut
+							 || INTERNAL_fadeMode == FadeMode.ReleaseRpc;
 			}
 		}
 
@@ -108,11 +109,10 @@ namespace Microsoft.Xna.Framework.Audio
 		private CueData INTERNAL_data;
 
 		// Current sound and its events
-		private XACTSound INTERNAL_activeSound;
-		private List<XACTEvent> INTERNAL_eventList;
-		private List<bool> INTERNAL_eventPlayed;
-		private Dictionary<XACTEvent, int> INTERNAL_eventLoops;
-		private Dictionary<SoundEffectInstance, XACTEvent> INTERNAL_waveEventSounds;
+		private XACTSoundInstance INTERNAL_activeSound;
+
+		private Dictionary<SoundEffectInstance, PlayWaveEventInstance>
+			INTERNAL_playWaveEventBySound;
 
 		// Used for event timestamps
 		private Stopwatch INTERNAL_timer;
@@ -120,10 +120,10 @@ namespace Microsoft.Xna.Framework.Audio
 		// Sound list
 		private List<SoundEffectInstance> INTERNAL_instancePool;
 		private List<double> INTERNAL_instanceVolumes;
-		private List<short> INTERNAL_instancePitches;
+		private List<float> INTERNAL_instancePitches;
 
 		// RPC data list
-		private List<float> INTERNAL_rpcTrackVolumes;
+		private List<double> INTERNAL_rpcTrackVolumes;
 		private List<float> INTERNAL_rpcTrackPitches;
 
 		// Events can control volume/pitch as well!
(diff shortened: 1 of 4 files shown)


flibitijibibo:
Will do my best to look at this on/by Tuesday, definitely before FNA 17.01. At first glance it doesn't change a whole lot, so at most it'll just need minor style fixes or something like that.

kiates:
I use ReSharper and after the struggle to merge after you applied style fixes before (since I had work in progress), I tried to configure it to match your style as much as I could.  If there is a pretty printer that you use and if there is a configuration available I'm happy to run it as a part of process before preparing pull-requests.

One thing I couldn't get ReSharper to do is get the ) on the end of calls to wrap to the next line.  Hopefully it's closer than previous pull-requests on other fronts.

flibitijibibo:
Sadly I'm nuts and do everything by hand, so I don't have any usable templates... I should probably crank up the motor on the Windows builder and put something together that ReSharper can use. Worst case, I'll deal with it all post-merge.

For now I'd just focus on making sure all the changes are intentional (I did see some rollbacks for the pitch/volume storage, wasn't sure if that was a merge conflict). As long as it functions, style's just glitter on top.

kiates:
Yes, what looks like rollbacks with pitch/volume in the parser are intentional.  In memory units are nominal Db and Semitones rather than 100s of each.  The 100s are only used as an artifact of the disk format, so as they are read into memory they are converted.  This is true for Cue, Sound, Track, Event volumes, as well as RPC values.  Xact seems to be consistent on this so it makes many functions operate at the same scaling factor.

flibitijibibo:
Would it be possible to separate the scaling changes from this patchset? Because some parts change existing storage while the new stuff assumes the new storage it makes things really hard to compare - once this is in we can check out the scaling on its own to see if it makes the code nicer. (Though tbh I'm inclined to say it won't; having now stepped through XACT with the untouched numbers at the end of INTERNAL_update it makes comparing calculations against XAP files super _super_ nice, so it would have to be a crazy big CPU/memory improvement for me to give that up.)

kiates:
I created a branch off of this one that uses the encoded values represented in the XACT file directly at run-time.  Though it keeps the float format for pitch rather than the short that is used in some places in the XACT file to hold pitch information.

src/Audio/Cue.cs
```diff
@@ -748,27 +828,26 @@ internal bool INTERNAL_update()
...
+				 * XACT uses -12 to 12 (+/- 12 semitones),
 				 * XNA uses -1.0f to 1.0f (+/- 1 octave).
 				 */
 				INTERNAL_instancePool[i].Pitch = (
 					INTERNAL_instancePitches[i] +
-					rpcPitch +
-					eventPitch +
-					INTERNAL_rpcTrackPitches[i]
-				) / 1200.0f;
+					(i == 0 ? rpcPitch : INTERNAL_rpcTrackPitches[i-1]) +
```
flibitijibibo: This is the only possible regression I can see - do you have sample data for the global RPC vs. track RPC selection? I always thought it was both, but I think my only sample data was Apotheon's music tracks.

flibitijibibo:
Finally got a chance to look over this, sorry it took so long.

Minus the scaling and the one comment, this all seems fine to me. That said, it's a lot of different things in one batch, so what I'll probably do since we're on a deadline is manually verify and merge each piece at a time on the 30th and 31st, and everything should be in for FNA 17.01 unless I find something wrong with one of the changes (and only that change will be omitted).

kiates:
I will decompose as much as possible into individual pull requests. You shouldn't need to perform the work just because I kept adding more and more features/fixes :)

src/Audio/Cue.cs
kiates: I created a dedicated pull request with changes related to this and an introduction as to the reasoning.

kiates: This pull request can be removed once we work through the the individual transformations represented in the others.  I was able to invest more time in the event refactor too, so it's much better and less of a functional work-in-progress.

flibitijibibo:
As of https://github.com/FNA-XNA/FNA/commit/259e0996423aff7564fe43bfeb12527350f85256 everything should be merged in. The only thing I really changed other than style things was the float/double and float/short changes, everything else should be the same. If not, let me know.

PR #95, opened 2016-12-27
Fix issue with rpc track volumes.

kiates:
Once I added support for Attack/Release RPCs I had a test case with volume attack and release targeting two separate tracks within a single sound.  The first track had an attack from -96db  up to 0db over 2 sec, with a release from 0db down to -96db over 2 seconds as well.  The second track was similar but the duration for both the attack and release was 8 seconds.

FACT did not behave correctly, instead it applied the 2 second attack/release to track 2 instead of track 1.

Looking through the code it appears that FACT uses rcpVolume for "simple" sounds, that is sounds with a single track.  Additional tracks (2 and up) are then stored as overflow in an array INTERNAL_rpcTrackVolumes[].  The code is littered with multiplex/demultiplex patterns to navigate this atypical data structure. Simplifying this is another endeavor, but it appeared that what was missing was the multiplexing of the RPC track volume across rpcVolume and INTERNAL_rpcTrackVolumes[].

Addition of this correctly produced the 2 sec attack/release on track 1 and the 8 sec attack/release on track 2.

I have not explicitly tested it, but by a similar argument, the application of pitch need the same operation.

### DIFF
--- a/src/Audio/Cue.cs
+++ b/src/Audio/Cue.cs
@@ -748,26 +748,25 @@ namespace Microsoft.Xna.Framework.Audio
 			for (int i = 0; i < INTERNAL_instancePool.Count; i += 1)
 			{
 				/* The final volume should be the combination of the
-				 * authored volume, category volume, RPC/Event volumes, and fade.
+				 * authored volume, category volume, RPC Track volume, 
+				 * Event volumes, and fade.
 				 */
 				INTERNAL_instancePool[i].Volume = XACTCalculator.CalculateAmplitudeRatio(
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Would it be easier if we moved rpcVolume into the array at 0 and got rid of the [- 1] part everywhere else? It'll avoid a guaranteed branch and hopefully dispose of any off-by-one errors in other places.

In a way, rpcVolume is basically for the 0th track, so it'd make it slightly more logical too...

flibitijibibo:
Fair enough! I'll merge this along with the other pieces when I'm back at the flibitTower.

PR #102, opened 2017-02-03
Rpc sound vs track

kiates:
This reverts changes made to volume/pitch calculations that weren't correctly keeping track and sound rpc targets straight.

RPC curves targeting sounds are interesting.  The primary reason, is that they are optional, meaning, if they are there they are stored in the 0 slot within the XACT data with tracks starting at 1 and on.  But if they are not there, then tracks start at 0 and on.  The tricky thing is you don't know what your dealing with during run-time processing and hence the trouble with the i and i-1's in the cue update method.  I've been able to find the identifiers somewhere on the net (can't remember where now) for what curve sets are contained.  Using this information I can keep a flag indicating whether a sound has RPCs targeting itself and then use this information at run-time to multiplex between sound/track.

For a test I made an RPC that would change the pitch based on the number of cue instances, but found that that variable was not yet handled.  Hence the second commit.  There are a few other tweaks that I'll indicate in the pull request.

### DIFF
--- a/src/Audio/Cue.cs
+++ b/src/Audio/Cue.cs
@@ -741,15 +741,24 @@ namespace Microsoft.Xna.Framework.Audio
 			float lfGain = 1.0f;
 			for (int i = 0; i < INTERNAL_activeSound.Sound.RPCCodes.Count; i += 1)
 			{
-				if (i > INTERNAL_instancePool.Count)
+				// Are we processing an RPC targeting the sound itself rather than a track?
+				bool isSoundRpc = i == 0 && INTERNAL_activeSound.Sound.HasSoundRpcs;
+
+				// If there is an RPC targeting the sound instance itself, it is handled in rpcVolume/rpcPitch, and the first track is at i-1.
+				int trackRpcIndex = INTERNAL_activeSound.Sound.HasSoundRpcs ? i - 1 : i;
+
+				// If this RPC Code is for a track that is not active yet, we have nothing to do.
+				if (trackRpcIndex >= INTERNAL_instancePool.Count)
 				{
+					// FIXME: This presumes that tracks start in order, which doesn't have to be true.
 					break;
 				}
-				if (i > 0)
+				if (!isSoundRpc)
 				{
-					INTERNAL_rpcTrackVolumes[i - 1] = 0.0f;
-					INTERNAL_rpcTrackPitches[i - 1] = 0.0f;
+					INTERNAL_rpcTrackVolumes[trackRpcIndex] = 0.0f;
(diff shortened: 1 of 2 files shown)


src/Audio/Cue.cs
```diff
@@ -774,6 +783,14 @@ internal bool INTERNAL_update()
 								long elapsedFromStop = INTERNAL_timer.ElapsedMilliseconds - INTERNAL_fadeStart;
 								variableValue = elapsedFromStop;
 							}
+							else
+							{
+								variableValue = 0.0f;
```
kiates: Since variableValue is unconditionally computed on line 768, this ensures that it is held at position 0 until it's time.

src/Audio/Cue.cs
```diff
@@ -774,6 +783,14 @@ internal bool INTERNAL_update()
...
+							}
+						}
+						else if (curRPC.Variable.Equals("NumCueInstances"))
```
flibitijibibo: Doesn't the GetVariable() call before this if block handle NumCueInstances already?

flibitijibibo:
Other than the NumCueInstances thing this all looks good to me!

src/Audio/Cue.cs
flibitijibibo: Nevermind, it was a bug in FNA!

https://github.com/FNA-XNA/FNA/commit/34ec74f48c286203aa27dc7c5a55b86e2eb8878d

kiates: At first I thought there was a missing SetVariable() somewhere, but when I went to plumb that in I ran into the public,private restrictions since it is a public API. For a bit I started factoring out private versions without those restrictions, but then found that NumCueInstances was being handled a bit differently, so just decided to keep things local here for now.

flibitijibibo: Makes sense, especially considering the special case I wrote wasn't even correct... >_<

Go ahead and remove this `else` block and I'll merge this all in.

flibitijibibo:
Merged with a bonus commit that copies the variableValue changes into a similar funciton, along with a micro-optimization:

https://github.com/FNA-XNA/FNA/commit/0b8047e1e7b56f5bee04f32e8cdc46bab295f62d

flibitijibibo:
Wow, this old thread... but as an FYI: I finally figured out how we're supposed to know what the RPC blocks are for - if FACT doesn't come out soon this could still be helpful for our existing implementation:

https://github.com/flibitijibibo/FACT/commit/63cfb24a3229836f8b501e1e5732d8a195e0b1ac

PR #108, opened 2017-03-08
Adds file drop support.

Nabile-Rahmani:
This adds an extension for dropping files into the game window, which could be handy.


flibitijibibo:
It definitely works, but I dunno if I'd make this an extension (and certainly not on Game.cs). At most I'd make it more like TextInputEXT where it's a separate static class, but unlike text input there are lots of important variables for DropEvents, like drop type, the type of text (which we would want UTF8 translation for, rather than ANSI) and target window.

You're better off using AddEventWatch in this case, even if it adds an SDL2 reference to your game. Good work though!

PR #111, opened 2017-03-15
[VideoPlayer] Replace TheoraPlay with Theorafile

flibitijibibo:
This is a bug tracker for the VideoPlayer rewrite using Theorafile.

The native libraries  for Theorafile can be downloaded here:

http://www.flibitijibibo.com/files/theorafile.tar.bz2

LAST UPDATE: March 16 17:00 EST

The intention is that this will remove a thread, make the player more accurate to WMP behavior, and dramatically reduce the amount of RAM used by the player (in some cases, by 2 orders of magnitude!).

The current test suite is as follows, if your game is not in this list then it is not currently being tested:

- PXTND
- Capsized
- AVNT
- Dust: AET
- Reus
- Fist Puncher
- Skulls of the Shogun
- Apotheon
- Cryptark
- Shuggy


flibitijibibo:
This will be going into mainline at noon EDT on Wednesday - a little under 48 hours from now. If you're using VideoPlayer in your game, be sure to test this before it goes out in 17.04!

PR #135, opened 2017-10-24
Add ContentReader.platform internal field; Attempt to convert X360 Tex2Ds

0x0ade:
This PR adds the `char ContentReader.PlatformEXT` property, which is the 4th "header" byte. It allows `ContentTypeReader`s to check the content platform and to handle the content accordingly.

It also adds basic support for `Color`, `ColorBgraExt` and `Dxt3` textures in X360 XNBs by swapping anything accordingly in `Texture2DReader` before `SetData`. This builds upon the above change.

### DIFF
--- a/FNA.Core.csproj
+++ b/FNA.Core.csproj
@@ -157,6 +157,7 @@
 		<Compile Include="src\Graphics\DirectionalLight.cs" />
 		<Compile Include="src\Graphics\DisplayMode.cs" />
 		<Compile Include="src\Graphics\DisplayModeCollection.cs" />
+		<Compile Include="src\Graphics\X360TexUtil.cs" />
 		<Compile Include="src\Graphics\DxtUtil.cs" />
 		<Compile Include="src\Graphics\Effect\Effect.cs" />
 		<Compile Include="src\Graphics\Effect\EffectAnnotation.cs" />
--- a/FNA.csproj
+++ b/FNA.csproj
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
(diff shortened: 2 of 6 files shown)


flibitijibibo:
PlatformEXT doesn't really need to be exposed as a public property, but that char can definitely be stored internally and used by the readers if need be.

For the Texture2DReader, try to design the Xbox path in a way that doesn't affect the existing path - what we can probably do is add a check alongside the `if (!S3TC)` checks to see if the platform is 'x', then we can do conversion and treat it as a modified format. The downside is that Xbox textures will never get the GetBuffer() optimization (though if you _really_ wanted to be fast you'd just bake the PC textures offline instead of converting at runtime, so :shrug: to whoever tries to ship that commercially), the upside is that this code (which only really gets used by XnaToFna and nobody else) doesn't need to get #ifdef'd for production builds since it only adds 1 jump. In its current form I'd want to add `#if XBOX_CONTENT` or something like that.

ContentManager's changes are fine, as is ContentReader's (aside from the extension).

0x0ade:
I've structured the X360 texture swapper similar to `DxtUtil` and reverted the MemoryStream behavior to keep things simple.

I didn't know what the utility class should be named, so I've named it `X360TexUtil` for now. I don't know if any further X360 content handling PRs will follow, but if so, I'll probably rename it to `X360ContentUtil` or similar.

flibitijibibo:
Changes look good to me. If you have any cleanup you want to do go ahead and do it (FNA.csproj diff being weird, etc.) and I'll merge this in.

0x0ade:
Would the `#if XBOX_CONTENT` conditional be placed in `Texture2DReader`? I don't know how "correct" it'd be there, as it would affect other content type readers, too (that is, if any further changes are required).

0x0ade:
Thanks for reminding me about the FNA.csproj changes, I forgot cleaning up after Visual Studio. Can be merged now.

flibitijibibo:
Any chance we can get Dxt1/Dxt5 swappers for FNA 17.11? Just want to make sure this feature is complete before advertising it.

0x0ade:
I'll look at it when everyone else is asleep at the jam, in a few hours. Unfortunately I don't have any Dxt1 / Dxt5 test data.

PR #136, opened 2017-10-28
Fix X360TexUtil fail if !SupportsS3tc and add (untested) X360 Dxt1 / Dxt5 swap

0x0ade:
It seems like Dxt1, Dxt3 and Dxt5 store their data in 16 bit "chunks" (at least Dxt3 does) and I actually _swapped the compressed data. It simply worked out of sheer luck._

This PR fixes the X360 Dxt3 swapper if `!reader.GraphicsDevice.GLDevice.SupportsS3tc` and adds Dxt1 and Dxt5 swappers, based on the assumption that all data is stored in 16 bit "chunks." 2 byte sized pairs need to be swapped in Dxt3.

This was only tested with X360 Dxt3 textures. I've thus added `SwapEndian` for `ushort`, `uint` and `ulong` to X360TexUtil as preparation if the "16 bit chunks" assumption turns out to be incorrect for Dxt1 / Dxt5.

I could've kept the existing Dxt3 swapping method and theoretically could've reused it for Dxt1 / Dxt5, but it didn't offer any sort of explanation on what's actually going on.

### DIFF
--- a/src/Content/ContentReaders/Texture2DReader.cs
+++ b/src/Content/ContentReaders/Texture2DReader.cs
@@ -136,11 +136,60 @@ namespace Microsoft.Xna.Framework.Content
 					continue;
 				}
 
+				// Swap the image data if required.
+				if (reader.platform == 'x')
+				{
+					if (	surfaceFormat == SurfaceFormat.Color ||
+						surfaceFormat == SurfaceFormat.ColorBgraEXT	)
+					{
+						if (levelData == null)
+						{
+							levelData = reader.ReadBytes(levelDataSizeInBytes);
+						}
+						levelData = X360TexUtil.SwapColor(
+							levelData
+						);
+					}
+					else if (surfaceFormat == SurfaceFormat.Dxt1		)
+					{
+						if (levelData == null)
+						{
+							levelData = reader.ReadBytes(levelDataSizeInBytes);
+						}
+						levelData = X360TexUtil.SwapDxt1(
+							levelData, width, height
+						);
+					}
+					else if (surfaceFormat == SurfaceFormat.Dxt3		)
+					{
+						if (levelData == null)
+						{
+							levelData = reader.ReadBytes(levelDataSizeInBytes);
+						}
+						levelData = X360TexUtil.SwapDxt3(
+							levelData, width, height
+						);
+					}
+					else if (surfaceFormat == SurfaceFormat.Dxt5		)
+					{
+						if (levelData == null)
+						{
+							levelData = reader.ReadBytes(levelDataSizeInBytes);
+						}
+						levelData = X360TexUtil.SwapDxt5(
+							levelData, width, height
+						);
+					}
+				}
+
 				// Convert the image data if required
 				if (	surfaceFormat == SurfaceFormat.Dxt1 &&
 					!reader.GraphicsDevice.GLDevice.SupportsDxt1	)
 				{
-						levelData = reader.ReadBytes(levelDataSizeInBytes);
+						if (levelData == null)
+						{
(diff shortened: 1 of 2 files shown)


flibitijibibo:
This may be helpful:

https://msdn.microsoft.com/en-us/library/windows/desktop/bb694531(v=vs.85).aspx

DXT1 is 8 bytes so that’ll be different for sure.

0x0ade:
> The BC2 format stores colors with the same number of bits and data layout as the BC1 format
> The BC3 format stores colors with the same number of bits and data layout as the BC1 format

If they store colors in the same data layout and given that DXT3 stores the color and lookup table as 4 shorts, DXT1 seems to consist of 4 shorts (effectively 8 bytes), too. I really wish I had actual data to test this on.

flibitijibibo:
May just have to use PC data and deliberately load it incorrectly :P

0x0ade:
I don't even have PC DXT1 data at hand 😅 oh dang...

Something that seems off to me is that according to the illustrations in the documentation, the `16 2-bit color indices` are split across 4 bytes - thus (by my understanding) they shouldn't be affected by the X360 being big-endian. But when swapping everything except the format table, the DXT3 fonts in "Death vs. The IRS" and "Mute Crimson" render incorrectly.

If I had to guess, Microsoft's documentation is incorrect with them reading / writing the table as two words / shorts instead of four bytes... or I'm doing something wrong.

image
image

0x0ade:
I just noticed that the DXT3 alpha data is already correct - I was _really_ lucky with the initial "fix" to this issue (swapping everything as if everything's a big-endian word / short)! The two colors and color table are still incorrect. Because of that, I've "reverted" the DXT5 and DXT3 alpha data manipulation, made `SwapDxt3Block` and `SwapDxt5Block` call `SwapDxt1Block` and squashed the commits. 

Out of interest regarding the DXT1 / BC1 format, I peeked at how the Xenia X360 emulator handles DXT1 / DXT3 / DXT5 textures - specifically at how libcrn ("crunch") handles the color index table. It stores two 16 bit halves ("low" and "high") in a "DXT1 block" struct, which it uses to compress / decompress DXT1, DXT3 and DXT5.

As it's the only functional X360 emulator to date and as it seems to be actually capable to read DXT1 textures, I'm starting to believe that the information on MSDN is incorrect - it seems to further confirm that the color table consists of two shorts / words, at least in X360 DXT-compressed textures.

flibitijibibo:
Will merge this for completion’s sake, but definitely try to get some test data - fabricate it from DDS images with GIMP if you have to.

Worst case I’ll just remove this, tag 17.11, and add it back in, but it’d be cool to start advertising XnaToFna’s WIP 360 stuff right away.

0x0ade:
I just got home, am finally able to find / fabricate test data. I thought about feeding DXT1 and DXT5 textures from non-XNA games into the bit swappers if I won't be able to find any XNBs containing DXT1 / DXT5 Tex2Ds.

By the way, congratulations!

0x0ade:
Got myself my hands on a copy of Terraria X360, which uses DXT1 and DXT5 textures. Then hooked up the XnaToFna XNB transformer to FNA with the updated X360TexUtil and made it dump all textures as PNGs.

`Content\Images\logo_1.xnb`, DXT1:
logo_1 xnb

This seems to work just fine.

`Content\Images\Cursor.xnb`, DXT5:
cursor xnb

I'll fix the DXT5 alpha transform ASAP. MSDN implies that alpha_0 and alpha_1 are stored as two separate bytes, which I'm adhering to right now. Also, I'm just copying over the 16 3-bit alpha values as-is.

PR #137, opened 2017-10-30
Xbox 360 DXT3 and DXT5 texture swappers fixed

0x0ade:
The DXT5 swapper is fixed, tested with the textures in Terraria X360. Also based on Terraria X360, the DXT1 swapper currently built into FNA is already fully functional.

Unrelated to the DXT5 swapper but related to my previous pull request, I also had to fix the DXT3 swapper one last time:

>I just noticed that the DXT3 alpha data is already correct - I was really lucky with the initial "fix" to this issue (swapping everything as if everything's a big-endian word / short)!

I wasn't at home during the weekend, so I only took the last snapped screenshots for comparison purposes and "Mute Crimson" for testing. That game looks okay when leaving the DXT3 alpha data as is.

But when I came back home yesterday, I tested the new DXT3 swapper on my main PC and noticed that the textures in other games were broken again.

I'm sorry for not having tested it enough before committing. I've now tested the current DXT3 swapper with all games I can throw at it right now and it seems to work properly now.

### DIFF
--- a/src/Graphics/X360TexUtil.cs
+++ b/src/Graphics/X360TexUtil.cs
@@ -163,43 +163,36 @@ namespace Microsoft.Xna.Framework.Graphics
 
 		private static void SwapDxt1Block(BinaryReader imageReader, BinaryWriter imageWriter)
 		{
-			// Swap 2 shorts.
+			// Fix the following two big-endian words to litte-endian words.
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Glad we could get this out for FNA 17.11, thanks for looking at it!

PR #138, opened 2017-11-02
Allow negative source width / height in SpriteBatch

0x0ade:
This replaces `max(size, epsilon)` (breaks with negative size) with `sign(size) * max(abs(size), epsilon)`. The performance impact should be minimal and it keeps divisions by zero away.


flibitijibibo:
Thanks!

One possible optimization is removing all the extra work for Draw()s where the PushSprite call does not use sourceW/sourceH as a divisor; PushSprite itself only uses it to multiply rather than divide so removing the work in those spots would be safe.

That's an extra thing though, will merge the fix as-is.

flibitijibibo:
Just did the optimization; it's limited to calls that do not have origin parameters:

https://github.com/FNA-XNA/FNA/commit/2f52bb3fc4e17e1986524bf607a44231458fcb82

PR #140, opened 2017-11-05
ExternalReferenceReader.TargetType is null in XNA

0x0ade:
It's "legal" for games to list / use `ExternalReferenceReader` more than once in a single XNB.

FNA fails because `contentReaders.Add(newReaders[i].TargetType, newReaders[i]);` in `ContentTypeReaderManager.LoadAssetReaders` fails if a mapping for the target type already exists.

While debugging the issue, I noticed that when using XNA, `ExternalReferenceReader.TargetType` is null and that `ContentTypeReaderManager.GetTypeReader(typeof(object))` doesn't return the reader. 

Sample code:
```cs
Type t_ExternalReferenceReader = typeof(ContentTypeReader).Assembly.GetType("Microsoft.Xna.Framework.Content.ExternalReferenceReader");
Type target = ((ContentTypeReader) Activator.CreateInstance(t_ExternalReferenceReader)).TargetType;
Console.WriteLine($"ExternalReferenceReader.TargetType: {target} (== null: {target == null})");
```

Output using XNA:
```
ExternalReferenceReader.TargetType:  (== null: True)
```

Output using FNA:
```
ExternalReferenceReader.TargetType: System.Object (== null: False)
```

### DIFF
--- a/src/Content/ContentReaders/ExternalReferenceReader.cs
+++ b/src/Content/ContentReaders/ExternalReferenceReader.cs
@@ -16,7 +16,7 @@ namespace Microsoft.Xna.Framework.Content
 	{
 		#region Public Constructor
 
-		public ExternalReferenceReader() : base(typeof(object))
+		public ExternalReferenceReader() : base(null)
 		{
 		}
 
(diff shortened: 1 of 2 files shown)


flibitijibibo:
`typeof(object)` does seem kind of vague, huh...

PR #151, opened 2018-02-08
Allow calls to Apply3D(AudioListener[] listeners, AudioEmitter emitte…

bartwe:
Allow calls to Apply3D(AudioListener[] listeners, AudioEmitter emitter) with no or 1 AudioListener

### DIFF
--- a/src/Audio/SoundEffectInstance.cs
+++ b/src/Audio/SoundEffectInstance.cs
@@ -263,6 +263,12 @@ namespace Microsoft.Xna.Framework.Audio
 
 		public void Apply3D(AudioListener[] listeners, AudioEmitter emitter)
 		{
+            if (listeners.Length == 0)
+                return;
+            if (listeners.Length == 1) {
+                Apply3D(listeners[0], emitter);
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Just needs style fixes and this is good to go:

Tabs, not spaces
Braces on all if/else blocks
Braces on the left

src/Audio/SoundEffectInstance.cs
```diff
@@ -263,6 +263,12 @@ public void Apply3D(AudioListener listener, AudioEmitter emitter)
 
 		public void Apply3D(AudioListener[] listeners, AudioEmitter emitter)
 		{
+            if (listeners.Length == 0)
```
flibitijibibo: I just took a second look and this actually isn't valid - if listeners is empty you get ArgumentOutOfRange. [1] is valid, however, so you can just take this first check out entirely.

bartwe:
Hope this is better :)

flibitijibibo:
Merged, thanks!

PR #155, opened 2018-03-05
Handle issue with memory stream internal array may have an _origin offset or might be larger than the memorystream itself

bartwe:
Handle issue with memory stream internal array may have an _origin offset or might be larger than the memorystream itself

### DIFF
--- a/src/Content/ContentReaders/Texture2DReader.cs
+++ b/src/Content/ContentReaders/Texture2DReader.cs
@@ -13,7 +13,7 @@
 
 #region Using Statements
 using System;
-
+using System.IO;
 using Microsoft.Xna.Framework.Graphics;
 #endregion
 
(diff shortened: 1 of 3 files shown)


flibitijibibo:
Just needs tabbing fixes and explicit variable type names, this is fine otherwise!

flibitijibibo:
Closing this as invalid. A couple reasons for this:

1. The X360Util changes don't make sense as the stream's internal buffer is in fact always going to be at 0, as the MemoryStream allocations in those functions will show. If you wanted to clean that up, you could create the `byte[]` as a separate variable, pass it to `new MemoryStream`, and return the variable directly at the end.
2. AFAIK the Seek return value is just Position, as the reference source shows:

https://referencesource.microsoft.com/#mscorlib/system/io/memorystream.cs

The second reason is harmless and the first reason adds an alloc for no reason. So there's nothing here we really want...

flibitijibibo:
On second look it seems I was looking at the wrong function for Position, I was looking at InternalPosition thinking it was used in the property later. So I'll just grab the origin for this... can do this in a separate patch.

flibitijibibo:
https://github.com/FNA-XNA/FNA/commit/177cdc60ed93160ad696e4c9288293d5f1f2a3dc

https://github.com/FNA-XNA/FNA/commit/db77b345f786a785f40c77711c2db905cbf6cb51

PR #160, opened 2018-04-07
fix openbsd and freebsd library names

thfrwn:
Sigh, these *BSD always coming up with their own rules ;-)

Explanation:

* OpenBSD doesn't use the `-2.0` for SDL2 and related libs
* major for libopenal and libvorbisfile that is available is different, but per testing compatible. The way I've been successfully working with it is to leave the major out of the config and let the dynamic linker link it to the available one. I know this is not ideal, but the best I can come up with in a situation where using bundled libraries is essentially out of the question.
* freebsd's libopenal major is listed as 0, would also suggest leaving out the version here
* I wasn't able to find `.so` in netbsd's plist. I don't run it myself, so I can't just look for an explanation. I've started asking around and if I find out how netbsd can be included, I'll let you know.

Ok?

### DIFF
--- a/FNA.dll.config
+++ b/FNA.dll.config
@@ -2,15 +2,18 @@
 <configuration>
 	<dllmap dll="SDL2.dll" os="windows" target="SDL2.dll"/>
 	<dllmap dll="SDL2.dll" os="osx" target="libSDL2-2.0.0.dylib"/>
-	<dllmap dll="SDL2.dll" os="linux,freebsd,openbsd,netbsd" target="libSDL2-2.0.so.0"/>
+	<dllmap dll="SDL2.dll" os="linux,freebsd,netbsd" target="libSDL2-2.0.so.0"/>
+	<dllmap dll="SDL2.dll" os="openbsd" target="libSDL2.so.0"/>
 
 	<dllmap dll="SDL2_image.dll" os="windows" target="SDL2_image.dll"/>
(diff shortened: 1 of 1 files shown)


flibitijibibo:
That all seems fine to me, but do we know why the OpenAL version numbering is still 0? AFAIK OpenAL Soft still supports all of those OSes natively, so are they using some old version at the moment?

thfrwn:
Did some more research and freebsd got an old port called [openal](https://www.freshports.org/audio/openal/) and apparently another called [openal-soft](https://www.freshports.org/audio/openal-soft/) which is version 1.18.2.

I'd just leave freebsd as it is then.

The versioning of openbsd's libs is primarily done in ports reportedly because of experience with upstream not handling major and minor bumps correctly. At least that would probably be the explanation for our openal 1.17 port installing a libopenal.so.3. libvorbisfile is version 6 which per the makefile is from 6.7 upstream. Sounds like it might be ahead of what's shipped with FNA, but noted no issues with it so far.

Anyhow, the current state of the PR should be suitable regarding openbsd. I'm hoping to hear from some other *BSD to make more informed decisions there...

flibitijibibo:
Sounds good, will go ahead and merge this. I wouldn't overthink this too much, as both OpenAL Soft and Vorbisfile are both getting replaced when FAudio is done. And I imagine the SDL2 changes aren't going to be undone any time soon... so whatevs. :shrug:

PR #161, opened 2018-04-16
Allow .NET 4.0 projects to reference FNA.Core

danielcrenna:
Multi-targeting will help where library developers want to support both variants in a single assembly.


flibitijibibo:
I had no idea this was possible, that’s pretty neat!

PR #167, opened 2018-05-09
Fix TitleLocation.Path regression

0x0ade:
Setting up `TitleLocation.Path` using a static constructor introduced a regression on games reflecting into `TitleLocation.Path` before` FNAPlatform` has been set up:

- Game reflects into `TitleLocation.Path`
- `TitleLocation` class constrctor calls `FNAPlatform.GetBaseDirectory`
- `FNAPlatform` class constructor runs
- `FNAPlatform` sets `GetBaseDirectory`
- `FNAPlatform` calls `ProgramInit`
- `ProgramInit` uses `TitleLocation.Path`
- `ProgramInit` receives uninitialized Path (== null)
- `ProgramInit` throws exception

The game in question is "Fusion: Genesis".

This "lazy property" fixes the issue while maintaining compatibility with all existing and future `TitleLocation.Path` usages.

### DIFF
--- a/src/TitleLocation.cs
+++ b/src/TitleLocation.cs
@@ -19,18 +19,29 @@ namespace Microsoft.Xna.Framework
 
 		public static string Path
 		{
-			get;
-			private set;
+			get
+			{
+				/* This property was previously prepared with the help
+				 * of a class (static) constructor and a private setter.
+				 * 
+				 * Unfortunately, it caused issues when the game reflected
+				 * into TitleLocation.Path before FNAPlatform had a chance
+				 * to initialize the platform, which itself requires the path.
+				 * 
+				 * This "lazy property" fixes the issue.
+				 * -ade
+				 */
+				if (INTERNAL_Path != null)
+					return INTERNAL_Path;
+				return INTERNAL_Path = FNAPlatform.GetBaseDirectory();
+			}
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Just to keep this a bit cleaner, does this work too?

```
get
{
    /* This has to be an explicit variable!
     * Reflection can query this before the static constructor is called,
     * and having the variable ensures this property is a function that
     * contains a call to the static constructor, so path is always initialized.
     */
    return path;
}

private static readonly string path = FNAPlatform.GetBaseDirectory();
```

flibitijibibo:
Oh wait I see the problem. Hm, give me a second while I unwrap this, I feel like I've made a horrible circular dependency here...

flibitijibibo:
Took a slightly different approach:

https://github.com/FNA-XNA/FNA/commit/4daadfcabc689294b5d21c0ecf78e0e0640dc637

This took more changes and might look a little weird on the FNAPlatform side, but it prevents a circular dependency which would make implementing a new FNAPlatform extremely confusing if you ran into this situation not knowing about it.

0x0ade:
Maybe having an `out string titleLocation` would make it look less weird, while also allowing you to expand it for other properties which require to be filled by the platform.

PR #168, opened 2018-05-10
Prevent propagating MediaPlayer's Volume property to already disposed Songs

advdotnet:
If you 

1. play a Song using MediaPlayer, 
2. stop this Song using MediaPlayer
3. dispose the Song / ContentManager
4. set the Volume property of MediaPlayer to any value

you get a NullReferenceException as the disposed song is still in the MediaPlayer's MediaQueue and the volume is propagated to every Song in the queue.

### DIFF
--- a/src/Media/Xiph/Song.cs
+++ b/src/Media/Xiph/Song.cs
@@ -112,7 +112,10 @@ namespace Microsoft.Xna.Framework.Media
 				 * We need to detach this source from the AL listener properties.
 				 * -flibit
 				 */
-				soundStream.Volume = value * (1.0f / SoundEffect.MasterVolume);
+				if (!IsDisposed) 
+				{
+					soundStream.Volume = value * (1.0f / SoundEffect.MasterVolume);
+				}
 			}
 		}
 


flibitijibibo:
Simple enough, though we should probably look at the MediaQueue to see if Disposing removes it in XNA.

flibitijibibo:
Here's a fun little discovery: The following is entirely valid in XNA:

```
Song one = Song.FromUri("xnaBlows", new Uri("test.wav", UriKind.Relative));
Console.WriteLine(one.Duration);
MediaPlayer.Play(one);
one.Dispose();
MediaQueue why = MediaPlayer.Queue;
Console.WriteLine(why.Count + " " + why.ActiveSong.Duration);
MediaPlayer.Volume = 0.5f;
```

Imagine what you think this does, and then read this list:
- The song does not stop playing at all
- Volume is set perfectly fine
- one's Duration is always TimeSpan.Zero regardless of format
- ActiveSong is not disposed and Duration is accurate!

So from what I can tell based on the locals, client-side Songs do almost nothing whatsoever and are almost completely ignored by the MediaPlayer. It's just a "handle" (which for all i know could literally just be a `const char*`) that MediaPlayer uses, so 100% of the meat of the player should probably be in that file.

The good news is the FAudio rewrite almost does this already, I just need to decouple the file handle from Song and the MediaPlayer's FAudio instance. Not difficult, just annoying...

PR #169, opened 2018-05-10
Add SetDataPointerEXT to VertexBuffer too

bartwe:


flibitijibibo:
Cool tabs.

Since this also affects IndexBuffer and can be moved up to the base class without consequence, I've done effectively this patch but across all the buffer types:

https://github.com/FNA-XNA/FNA/commit/ad77c06662db774df2750bbc631cdf856fd18e8c

PR #170, opened 2018-05-11
Use DeadSounds as a queue for sounds that need to be stopped.

bartwe:
Works around ConcurrentModificationException
Works around sounds getting stopped every update tick after first time the sound got added.
(Is there an issue here that if two copies of the sound are started with a few seconds offset, that stopping one doesn't cancel the other?)

### DIFF
--- a/src/FrameworkDispatcher.cs
+++ b/src/FrameworkDispatcher.cs
@@ -34,10 +34,14 @@ namespace Microsoft.Xna.Framework
 			/* Updates the status of various framework components
 			 * (such as power state and media), and raises related events.
 			 */
-			foreach (SoundEffectInstance sound in DeadSounds)
-			{
-				sound.Stop(true);
-			}
+            while (true) {
+                var count = DeadSounds.Count;
+                if (count == 0)
+                    break;
+                count--;
+                DeadSounds[count].Stop(true);
+                DeadSounds.RemoveAt(count);
+            }
 			foreach (DynamicSoundEffectInstance stream in Streams)
 			{
 				stream.Update();


flibitijibibo:
Just to avoid the `while(1)`, does this also work?

```
/* DeadSounds gets sent instances by the FAudio thread, so this can NOT
 * be a foreach! But on the bright side, DeadSounds only gets Adds from
 * the callback and only we Remove, so we can just keep removing [0].
 */
while (DeadSounds.Count > 0)
{
    DeadSounds[0].Stop(true);
    DeadSounds.RemoveAt(0);
}
```

bartwe:
It would, also makes the behavior more queue like.
Reason i did it from the back is that RemoveAt(0) is O(n) RemoveAt(Count-1) is O(1)
Should just use the Queue<T> type i guess
Also if the FAudio thread does the Adds, this is all still wrong, failures are just less likely to be detected but will still happen.

flibitijibibo:
We can use ConcurrentQueue instead - the audio thread will get priority and Dispatcher can use TryDequeue, so the thread doesn’t wait and Dispatcher can catch up later.

bartwe:
Given that it is an audio thread, probably best to use SpinLock so that the thread doesn't get yielded on contention, contention is only a few dozen instructions anyways. Except for the Queue which might do an alloc, but that is on the audio thread, so aslong as no gc happens that is fine.

flibitijibibo:
Go ahead and try the various ideas and see which one works best. Whatever can get you looking at the X3DAudio stuff which is actually on a deadline, unlike this.

flibitijibibo:
Someone else is looking at the `fact` branch and probably won't discover this so for now I've at least fixed the `foreach` contention:

https://github.com/FNA-XNA/FNA/commit/16922abc669a76b284c218b9e326892a80faf212

You can optimize this if you want, and you can use this pull request to do so.

bartwe:
Followed up in https://github.com/FNA-XNA/FNA/pull/174

PR #171, opened 2018-05-11
Added IsUnclampedEXT property to SoundEffectInstance

Enichan:
Using OpenAL there is no real reason why FNA needs to clamp pitch for sounds, aside from compatibility with XNA. This adds an extension to SoundEffectInstance called IsUnclampedEXT, which defaults to false and when set to true bypasses the clamping done in SoundEffectInstance.Pitch's setter and makes sure to set the clamp argument for AudioDevice.ALDevice.SetSourcePitch to false.

### DIFF
--- a/src/Audio/SoundEffectInstance.cs
+++ b/src/Audio/SoundEffectInstance.cs
@@ -73,13 +73,15 @@ namespace Microsoft.Xna.Framework.Audio
 			}
 			set
 			{
-				if (INTERNAL_isXACTSource)
-				{
-					value = MathHelper.Clamp(value, -2.0f, 2.0f);
-				}
-				else
-				{
-					value = MathHelper.Clamp(value, -1.0f, 1.0f);
+				if (!IsUnclampedEXT) {
+					if (INTERNAL_isXACTSource)
+					{
+						value = MathHelper.Clamp(value, -2.0f, 2.0f);
+					}
+					else
+					{
+						value = MathHelper.Clamp(value, -1.0f, 1.0f);
+					}
 				}
 				INTERNAL_pitch = value;
 				if (INTERNAL_alSource != null)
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Unfortunately this won’t work with the rewrite - you’ll still have to abide by the XAudio2 max frequency ratio as defined by both the global spec and the max dictated by the source voice at creation time, that’s not trivial to customize:

https://msdn.microsoft.com/en-us/library/windows/desktop/microsoft.directx_sdk.ixaudio2.ixaudio2.createsourcevoice(v=vs.85).aspx

I would have a look at using FAudio directly - you can probably come up with a far superior API than what Microsoft came up with, tbh.

Enichan:
You'll have to forgive me for asking but, what's FAudio? Also does this mean you're going to get rid of OpenAL entirely or something?

flibitijibibo:
Yup, replacing AL with a plain C XAudio2 implementation:

https://github.com/FNA-XNA/FNA/issues/90

https://github.com/flibitijibibo/FACT

Enichan:
So looking at that unless I'm misreading it doesn't it say you can go 10 octaves up or down? Wouldn't that still be a good use case for this extension then?

flibitijibibo:
Not quite - you would need to rework this to dictate the max at creation time, and part of the reason they want you to specify a max is because it’s how they determine the size of the decode/resample caches. If you generically apply a high max across your data you risk making the cache really big when something comes along with a high sample rate.

To best understand this, I recommend trying your idea out with the `fact` branch. It should have a new submodule, so be sure to init/update the submodules when you check it out. Then FNA should build and the new FAudio library can be built from here (prebuilts will be added when the rewrite is done):

https://github.com/flibitijibibo/FACT/tree/master/visualc

flibitijibibo:
Closing since this will get steamrolled over by the rewrite, you’ll need to either rework this for the `fact` branch or use FAudio directly. OpenAL# can also be used directly but note that you’ll be tied to the AL Soft license if you opt for that path. MojoAL hasn’t been tested with AL# yet but is a possibility if you use an extremely small subset of the AL spec.

Issue #172, opened 2018-05-12
Is it possible to capture the details of ARB_debug_output in an exception?


mklingen:
Our application has automatic crash reporting based on exception handling. Occasionally, we get exceptions reported like this one: https://github.com/CompletelyFairGames/dwarfcorp/issues/706

Such exceptions are difficult to debug because the actual content of the warning or error seems to just get `printfed` to the console (or in our case, a local log file). Is there a way to get the actual content of the warning when the exception is created? Or is there a way to capture such warnings when they occur so we can report them?

flibitijibibo:
I wouldn’t be opposed to making the Exception print the message too - just add it to both OpenGLDevice_GL and ModernGLDevice_GL and I can merge it in.

mklingen:
Cool, will make a PR when I get the chance.

PR #173, opened 2018-05-12
throw the actual error text inside OpenGLDevice for DEBUG_TYPE_ERROR_ARB

mklingen:
#172


flibitijibibo:
Quick and easy, thanks much!

PR #174, opened 2018-05-13
Make threadsafe FrameworkDispatcher.EnqueueSoundEffectInstanceStop

bartwe:
Use SpinLock  to protect the DeadSounds when accessing from multiple threads.
Using a SpinLock to avoid yielding the very latency sensitive mixer thread.

### DIFF
--- a/src/Audio/SoundEffectInstance.cs
+++ b/src/Audio/SoundEffectInstance.cs
@@ -481,7 +481,7 @@ namespace Microsoft.Xna.Framework.Audio
 		{
 			FAudio.FAudioSourceVoice_Stop(handle, 0, 0);
 			State = SoundState.Stopped;
-			FrameworkDispatcher.DeadSounds.Enqueue(this);
+			FrameworkDispatcher.EnqueueSoundEffectInstanceStop(this);
 		}
 
 		private unsafe void SetPanMatrixCoefficients()
--- a/src/FrameworkDispatcher.cs
+++ b/src/FrameworkDispatcher.cs
@@ -9,7 +9,7 @@
 
 #region Using Statements
 using System.Collections.Generic;
-
+using System.Threading;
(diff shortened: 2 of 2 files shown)


flibitijibibo:
Not thrilled about the `while(1)` but will test when I can sit down with it. If you can get rid of that I’d be willing to merge it right away.

bartwe:
You need the count check and dequeue call paired together in the same lock so that they are in sync, and you don't want to hold the lock during the stop so this seemed like the most obvious way of doing it.

flibitijibibo:
A good way to deal with the while(1) is use TryEnter on Update - the try/catch blocks are really slow on both threads; the Enqueue call should use Enter and the Dequeue call should use TryEnter and move on if it fails. try/catch are slow and Enter can stall, so this will just make things slow in addition to looking ugly (tabbing aside).

bartwe:
The try/finally are merely for correctness in the context of out of memory situations.
the TryEnter is only applicable to the dequeue, where we don't care about contention and the thread doesn't yield but spins so it barely matters. try/finally isn't slow but it is verbose

flibitijibibo:
I'd rather not slow programs down to account for OoM situations - by the time you hit that, you're screwed anyway (as I'm sure you're aware). What may be helpful is to allocate a reasonable size for the Queue right away, so Enqueues don't necessarily have to alloc when a handful of sounds die together. I'd be up for preallocating ~16-32; if more than 32 sounds all die at once something scary is probably happening...

Just for laughs, let's go ahead and add that TryEnter. If it really doesn't matter then worst case nothing will happen, and best case I can have peace of mind regarding the main thread stalling.

flibitijibibo:
Closing due to old age - give this another go when you've got time; strangely my SFI test case is totally fine with what exists now. My main thing is XACT though, so it hasn't had a lot of time to just idle or what have you.

PR #176, opened 2018-05-17
Fix MemoryStream buffer privacy ambiguity

0x0ade:
FNA uses `MemoryStream.GetBuffer`, which causes issues when a game's `ContentManager.OpenStream` returns a `MemoryStream` with a private buffer.

This pull request changes the behavior in `Texture2DReader` and `Texture2D` to try `MemoryStream.GetBuffer` first. If it fails, it falls back to reading the stream's contents.

In `Texture2DReader` specifically, all codepaths now lead up to

```cs
texture.SetData(level, null, levelData, levelDataByteOffset, levelDataSizeInBytes);
```

with the `ReadBytes` fallback being used only if no conversion occurred and if the buffer wasn't successfully obtained beforehand.

`Texture2DReader` and `Texture2D` were the only places I've found where `MemoryStream.GetBuffer` is being used.

Edit: To clarify, the "ambiguity" (misassumption? I don't know what word fits this best 😅) stems from the fact that we always assumed that the passed `MemoryStream`'s buffer is public. While this is guaranteed when we're internally using a `MemoryStream` when f.e. decompressing, this assumption is incorrect for `MemoryStream`s provided by the game.

### DIFF
--- a/FNA.Core.csproj
+++ b/FNA.Core.csproj
@@ -329,6 +329,7 @@
 		<Compile Include="src\TitleLocation.cs" />
 		<Compile Include="src\Utilities\AssemblyHelper.cs" />
 		<Compile Include="src\Utilities\FileHelpers.cs" />
+		<Compile Include="src\Utilities\FNAInternalExtensions.cs" />
 		<Compile Include="src\Vector2.cs" />
 		<Compile Include="src\Vector3.cs" />
 		<Compile Include="src\Vector4.cs" />
--- a/FNA.csproj
+++ b/FNA.csproj
@@ -397,6 +397,7 @@
     <Compile Include="src\TitleLocation.cs" />
     <Compile Include="src\Utilities\AssemblyHelper.cs" />
     <Compile Include="src\Utilities\FileHelpers.cs" />
+    <Compile Include="src\Utilities\FNAInternalExtensions.cs" />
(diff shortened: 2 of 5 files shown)


flibitijibibo:
Would it be worth exploring a check that digs into MemoryStream's private variables? I feel like at this point MemoryStream's access variable is uniform across .NET/Core/Mono at this point, but I could be wrong on that (even then it's easy to go down a list). The main concern is that wrapping this in a try block can put a pretty big dent into performance when the GetBuffer path was designed to be as speedy as possible. I'd be up for making a static FieldInfo somewhere that lets us pull this bool out from wherever we're working.

0x0ade:
I've added a `TryGetBuffer` extension method, which behaves similarly to the .NET 4.6 `TryGetBuffer` method. The only difference is that the `TryGetBuffer` extension provides a `byte[]` instead of an `ArraySegment<byte>`. 

As I didn't see any place where this would be best fit, I've created a new `internal static class FNAInternalExtensions` inside the `Utilities` directory, housing the extension method with the related reflection field info.

flibitijibibo:
Finally merged this, sorry for the wait. Also did some much needed cleanup of all the junk we do in the middle between reading and uploading:

https://github.com/FNA-XNA/FNA/commit/4d5de7cac57524cbc497fed73dc29fbb1518faee

PR #178, opened 2018-06-01
Make parameters in operator overloads match

WardBenjamin:
This also fixes some mismatched documentation for the `operator /(Vector3 value, float divider)` method.

### DIFF
--- a/src/Vector3.cs
+++ b/src/Vector3.cs
@@ -1411,12 +1411,12 @@ namespace Microsoft.Xna.Framework
 		/// <summary>
 		/// Inverts values in the specified <see cref="Vector3"/>.
 		/// </summary>
-		/// <param name="value">Source <see cref="Vector3"/> on the right of the sub sign.</param>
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Ignoring that this doesn't compile, it also doesn't match the spec. MSDN references, in order:

https://msdn.microsoft.com/en-us/library/microsoft.xna.framework.vector3.op_unarynegation(v=xnagamestudio.40).aspx
https://msdn.microsoft.com/en-us/library/bb198113(v=xnagamestudio.40).aspx
https://msdn.microsoft.com/en-us/library/bb198119(v=xnagamestudio.40).aspx
https://msdn.microsoft.com/en-us/library/bb196368(v=xnagamestudio.40).aspx

WardBenjamin:
Sorry about that, my bad. I've opened #179 with just the mismatched documentation fix, and verified that it matches the XNA MSDN docs.

PR #182, opened 2018-06-27
Add Texture2D.SetDataPointerEXT like on the VertexBuffer

bartwe:


src/Graphics/Texture2D.cs
```diff
@@ -158,7 +158,7 @@ int elementCount
 				w = Math.Max(Width >> level, 1);
 				h = Math.Max(Height >> level, 1);
 			}
-
+			var elementSize = Marshal.SizeOf(typeof(T));
```
flibitijibibo: Only one `var`, too! Aside from this the patch is good, so just rinse and repeat for Get and 3D/Cube.

bartwe:
Added the additional changes and removed the 'var'

Could not change IGLDevice due to the implementations requiring these fields for the other functions which also leaves the SetDataPointerEXT  more verbose in those cases

flibitijibibo:
For GetData it looks like you're right; even for buffer objects I opted to stick with the element stuff... SetData, however, I'm not seeing what'd make it more verbose. The GL impls wouldn't use `dataLength`, for example, seeing that they don't use `elementCount`, but VulkanDevice will need it for the memcpy operation and the pointer offsets that are calculated in the Devices can be done by the caller.

So I guess we can leave GetData alone (so that extension can be thrown out, sorry about that!) but SetDataPointer can definitely be applied to all three types.

bartwe:
Ok so what do you want me to do with the PR ? remove the GetData ones  and
further nothing ?

src/Graphics/Texture3D.cs
```diff
@@ -129,14 +130,42 @@ int elementCount
...
+			int left,
+			int top,
+			int right,
+			int bottom,
+			int front,
+			int back,
+			IntPtr data,
+			int dataLength
+		) {
+			if (data == null) {
```
flibitijibibo: Just this one brace, otherwise this is now set to go!

bartwe:
My autoformatter hates your autoformatter ;) do you maybe just have a
resharper autoformatter template i can import so it does this automatically
instead of manually ?

flibitijibibo:
I don't have any autoformatting locally, just autotabbing, so unfortunately I don't have any templates for anything.

bartwe:
pushed whitespace change

flibitijibibo:
[Docs have been updated.](https://github.com/FNA-XNA/FNA/wiki/5:-FNA-Extensions#setdatapointerext)

PR #183, opened 2018-06-27
Do not leak a string for every keypress.

bartwe:
Also combine Contains/Get pairs into TryGetValue calls

### DIFF
--- a/src/FNAPlatform/SDL2_FNAPlatform.cs
+++ b/src/FNAPlatform/SDL2_FNAPlatform.cs
@@ -722,7 +722,8 @@ namespace Microsoft.Xna.Framework
 					else if (evt.type == SDL.SDL_EventType.SDL_KEYUP)
 					{
 						Keys key = ToXNAKey(ref evt.key.keysym);
-						if (keys.Remove(key)) {
+						if (keys.Remove(key))
+						{
 							int value;
 							if (textInputBindings.TryGetValue(key, out value))
 							{
@@ -854,13 +855,14 @@ namespace Microsoft.Xna.Framework
 								while (*endPtr != 0)
 								{
 									endPtr++;
-									bytes++;
+									bytes += 1;
 								}
-								// utf8 will never encode more characters than bytes in a string so bytes is a suitable upper estimate of buffersize needed
+								// utf8 will never encode more characters than bytes in a string 
+								// so bytes is a suitable upper estimate of buffersize needed
 								char* charsBuffer = stackalloc char[bytes];
 								int chars = Encoding.UTF8.GetChars(evt.text.text, bytes, charsBuffer, bytes);
 
-								for (int i = 0; i<chars; ++i)
+								for (int i = 0; i < chars; ++i)
 								{
 								   TextInputEXT.OnTextInput(charsBuffer[i]);
 								}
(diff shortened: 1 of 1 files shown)


src/FNAPlatform/SDL2_FNAPlatform.cs
```diff
@@ -722,11 +722,11 @@ public static void RunLoop(Game game)
 					else if (evt.type == SDL.SDL_EventType.SDL_KEYUP)
 					{
 						Keys key = ToXNAKey(ref evt.key.keysym);
-						if (keys.Remove(key))
-						{
-							if (textInputBindings.ContainsKey(key))
+						if (keys.Remove(key)) {
```
flibitijibibo: Brace...

src/FNAPlatform/SDL2_FNAPlatform.cs
```diff
@@ -844,24 +844,27 @@ public static void RunLoop(Game game)
...
+							if (*endPtr != 0)
 							{
-								endPtr++;
+								int bytes = 0;
+								while (*endPtr != 0)
+								{
+									endPtr++;
+									bytes++;
+								}
+								// utf8 will never encode more characters than bytes in a string so bytes is a suitable upper estimate of buffersize needed
```
flibitijibibo: Long line...

src/FNAPlatform/SDL2_FNAPlatform.cs
```diff
@@ -844,24 +844,27 @@ public static void RunLoop(Game game)
...
+								char* charsBuffer = stackalloc char[bytes];
```
flibitijibibo: Not NET40 compatible, maybe guard this with NETSTANDARD2_0 like we do with SDL2#:

https://github.com/flibitijibibo/SDL2-CS/blob/master/src/SDL2.cs#L84

src/FNAPlatform/SDL2_FNAPlatform.cs
```diff
@@ -844,24 +844,27 @@ public static void RunLoop(Game game)
...
+								int chars = Encoding.UTF8.GetChars(evt.text.text, bytes, charsBuffer, bytes);
+
+								for (int i = 0; i<chars; ++i)
```
flibitijibibo: This probably has to get guarded too, and the NET40 compatible version can stick with the original `foreach`

flibitijibibo:
Other than the comments this is fine.

src/FNAPlatform/SDL2_FNAPlatform.cs
flibitijibibo: C# spec is written poorly and implies this requires modern C# when it doesn't, so it turns out this might be okay, I think. May have to test with BRUTE and Mono first.

src/FNAPlatform/SDL2_FNAPlatform.cs
flibitijibibo: (Also spacing and use of `++` instead of `+= 1`

flibitijibibo:
Let me go try this with BRUTE real quick, if it works there I'll go ahead and merge this

flibitijibibo:
Applied this to the libs as well, this was a good find!

https://github.com/FNA-XNA/FNA/commit/375203f59e6b00bf42399f77938a5e94eb66de41

bartwe:
Change to libs required,

chars isn't zero terminated after this, so explicit length needs to be
passed

int len = (int) (ptr - (byte*) s);
char* chars = stackalloc char[len];
int charsLength = System.Text.Encoding.UTF8.GetChars((byte*) s, len, chars,
len);
string result = new string(chars, 0, charsLength);

PR #184, opened 2018-06-27
Add SoundEffect NewSoundEffectPointerEXT comparable to Texture SetDat…

bartwe:
Never copy the SoundEffect array when called with an offset, expose NewSoundEffectPointerEXT

### DIFF
--- a/src/Audio/AudioDevice.cs
+++ b/src/Audio/AudioDevice.cs
@@ -11,6 +11,8 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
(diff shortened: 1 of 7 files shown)


flibitijibibo:
Rejecting since FAudio is bulldozing over this, and the way FAudio works this extension would be potentially redundant...?

https://github.com/FNA-XNA/FNA/blob/fact/src/Audio/SoundEffect.cs#L184

No reason this patch wouldn't work though, but I'm pretty much deprecating the old Audio system at this point.

bartwe:
K fair, i'll build a FACT specific version

PR #187, opened 2018-06-27
Replace IGLBuffer and the derived classes with GLBuffer struct to avo…

bartwe:
…id allocations

(cherry picked from commit e5a272ccdd9ef5015181f4268312a3eecbedaf7d)


flibitijibibo:
I’m willing to believe some noise comes from the type casts but at best this is a micro optimization (one that even our low-end targets don’t notice) and at worst it’s a _terrible_ way to fatally poison the abstraction layer. How am I supposed to work on a VulkanDevice with this?!

Come up with a better way to make an opaque type; I’m up for replacing the use of interfaces but this is unquestionably a net negative. I’d take IntPtrs even.

bartwe:
Fair, wasn't looking at the vulkan angle.
I don't care about the type casts, any allocation, can cause a garbage collection, with a duration proportional to the heap size, our heap is huge and causes significant spike frames/framedrops. Any individual allocation is tiny, but it has a massive impact on how smooth the game feels to not cause a gc periodically.

PR #204, opened 2018-10-22
Add Touch Support

TheSpydog:
This PR implements the `Microsoft.Xna.Framework.Input.Touch` namespace, as per #2.

Big highlights include:
- Accuracy fixes for the existing Touch stub
- Updates to FNAPlatform to detect SDL2 touch events and fetch device capabilities
- TouchPanel now registers up to 8 touches, which can be accessed via `TouchPanel.GetState()`
- Added an internal `GestureDetector` class whose sole responsibility is (you guessed it!) detecting Gestures
- All WP7 gestures are supported: Tap, DoubleTap, FreeDrag, HorizontalDrag, VerticalDrag, DragComplete, Flick, Pinch, and PinchComplete -- the gang's all here!
- Supports Windows desktop, unlike XNA!

After testing extensively with Nick Gravelyn's [XNA Gesture Sample](https://blogs.msdn.microsoft.com/nicgrave/2010/07/12/touch-gestures-on-windows-phone-7/), as well as a few other XNA sample games, I believe it works pretty well...at least on my Windows 10 machine! This will definitely need to be tested on different devices (and different touch-enabled OS's if possible).

### DIFF
--- a/src/FNAPlatform/SDL2_FNAPlatform.cs
+++ b/src/FNAPlatform/SDL2_FNAPlatform.cs
@@ -37,8 +37,6 @@ namespace Microsoft.Xna.Framework
 
 		private static bool SupportsGlobalMouse;
 
-		private static bool TouchDeviceExists;
-
 		private static float HapticMaxWorkaround;
 
 		#endregion
@@ -731,7 +729,7 @@ namespace Microsoft.Xna.Framework
 			}
 
 			// Perform initial check for a touch device
-			TouchDeviceExists = GetTouchCapabilities().IsConnected;
+			TouchPanel.TouchDeviceExists = GetTouchCapabilities().IsConnected;
 
 			// Do we want to read keycodes or scancodes?
 			if (UseScancodes)
@@ -831,7 +829,7 @@ namespace Microsoft.Xna.Framework
 					else if (evt.type == SDL.SDL_EventType.SDL_FINGERDOWN)
 					{
 						// Windows only notices a touch screen once it's touched
-						TouchDeviceExists = true;
+						TouchPanel.TouchDeviceExists = true;
 
 						TouchPanel.INTERNAL_onTouchEvent(
 							(int)evt.tfinger.fingerId,
@@ -1019,10 +1017,6 @@ namespace Microsoft.Xna.Framework
 				}
 
 				Keyboard.SetKeys(keys);
-				if (TouchDeviceExists)
-				{
-					TouchPanel.INTERNAL_updateTouches();
-				}
 				game.Tick();
 			}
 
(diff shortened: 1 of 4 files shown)


src/FNAPlatform/SDL2_FNAPlatform.cs
```diff
@@ -970,6 +1011,10 @@ public static void RunLoop(Game game)
 				}
 
 				Keyboard.SetKeys(keys);
+				if (GetTouchCapabilities().IsConnected)
```
flibitijibibo: This adds an allocation per frame... may just have to check for init, or call SDL_GetNumTouchDevices directly. I'd say check for touch on FNAPlatform.Init but that Windows thing is killing us...

src/Input/Touch/GestureDetector.cs
```diff
@@ -0,0 +1,640 @@
+﻿using System;
```
flibitijibibo: `#region License` at the top of the file, and `#region Using Statements` for these lines here.

flibitijibibo:
A whole bunch of comments as seen above, but man does this look nicer than the old Touch stuff. Tomorrow morning I'll give this a shot and provide feedback.

This is a total shot in the dark, but @nickgravelyn, if you have a PC with touch support, would you be up for trying this out to see how it feels? @TheSpydog has a prepared FNA sample to make testing this easier (can you post that for us, Caleb?).

nickgravelyn:
I do not have a PC with touch support, I'm afraid.

flibitijibibo:
Well, shoot - would have been cool to bring XNA Touch support full circle like that!

Another person that will need to see this is @0x0ade. Maik, see if this Caleb's FNA sample with FNADroid - functionally it seems to be the same as far as event handling goes, so I'm hoping this should be okay right off the bat.

flibitijibibo:
Well, this has certainly passed my test:

https://twitter.com/flibitijibibo/status/1054522305118904320

@0x0ade, if you think you'll have time within the week to play with this let me know, otherwise I'll do one last scan and merge this in. The binaries in the sample should load up without any fuss, I hope.

src/FNAPlatform/SDL2_FNAPlatform.cs
```diff
@@ -970,6 +1019,10 @@ public static void RunLoop(Game game)
 				}
 
 				Keyboard.SetKeys(keys);
+				if (TouchDeviceExists)
```
flibitijibibo: I just realized: This part may actually need to go into FrameworkDispatcher instead, for accuracy:

https://github.com/FNA-XNA/FNA/blob/master/src/FrameworkDispatcher.cs#L31

Functionally it's not a big deal for this one block, but the catch is that `TouchDeviceExists` may have to become an internal variable in TouchPanel or something... the init process can be the same, it's just this one update that has to move.

0x0ade:
I'll have some time to test this on Android either tomorrow or on Thursday.

PR #205, opened 2018-10-22
CoreRT AOT support

BurgerBob:
Added UnmanagedFunctionPointerAttribute to GL delegates
Better error report if GL loading fails.
rd.xml for reference

### DIFF
--- a/src/FNAPlatform/ModernGLDevice_GL.cs
+++ b/src/FNAPlatform/ModernGLDevice_GL.cs
@@ -217,11 +217,11 @@ namespace Microsoft.Xna.Framework.Graphics
 			GL_DEBUG_SEVERITY_NOTIFICATION_ARB =	0x826B
 		}
 
-		// Entry Points
+        // Entry Points
 
-		/* BEGIN GET FUNCTIONS */
+        /* BEGIN GET FUNCTIONS */
 
-		private delegate IntPtr GetString(GLenum pname);
+        [UnmanagedFunctionPointer(CallingConvention.StdCall)]private delegate IntPtr GetString(GLenum pname);
 		private GetString INTERNAL_glGetString;
 		private string glGetString(GLenum pname)
 		{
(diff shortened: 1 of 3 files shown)


src/FNAPlatform/ModernGLDevice_GL.cs
```diff
@@ -217,11 +217,11 @@ internal enum GLenum : int
 			GL_DEBUG_SEVERITY_NOTIFICATION_ARB =	0x826B
 		}
 
-		// Entry Points
+        // Entry Points
```
flibitijibibo: Tabbing's busted for the first few lines of changes. Also, please put the [Attribute] one line above the delegate declaration.

src/FNAPlatform/OpenGLDevice_GL.cs
```diff
@@ -855,7 +855,8 @@ bool enabled
 			if (type == GLenum.GL_DEBUG_TYPE_ERROR)
 			{
 				FNALoggerEXT.LogError(err);
-				throw new InvalidOperationException(err);
+                //On a des exceptions ici � cause de l'overlay steam. Si on commente le throw, �a marche quand m�me.
```
flibitijibibo: Looks like a temp comment got left in...

flibitijibibo:
Googling `rd.xml` really quickly, I'm thinking the config is going to need a whooole lot more entries than what we have right now. For now you can keep that file local to your FNA install, and if this becomes a widespread problem we can revisit that as a bundled file. All the CoreRT stuff is 100% new to me so I need to research this first before I start putting in dedicated commits.

Aside from that and the above comments, this LGTM.

src/FNAPlatform/OpenGLDevice_GL.cs
BurgerBob: Oh yes I forgot it. There is a throw here if the Steam Overlay is on, but it can be commented and the game still works.

flibitijibibo:
Ended up doing a search-and-replace to fix this: https://github.com/FNA-XNA/FNA/commit/8e4a826af519532a593212d23f928ebd487e0958

PR #206, opened 2018-11-01
Added rewind to the start of the stream at the Texture2D.FromStream

rds1983:
Right now, following code doesnt work in the FNA:
```c#
            using (var memoryStream = new MemoryStream())
            {
                using (var fileStream = File.OpenRead("d:\\temp\\default_uiskin.png"))
                {
                    fileStream.CopyTo(memoryStream);
                }

                _texture = Texture2D.FromStream(GraphicsDevice, memoryStream);
            }
```
Because memoryStream points to the end just before Texture2D.FromStream call.
While same code works well in the XNA.
This PR makes FNA work same as XNA.

### DIFF
--- a/src/Graphics/Texture2D.cs
+++ b/src/Graphics/Texture2D.cs
@@ -337,6 +337,11 @@ namespace Microsoft.Xna.Framework.Graphics
 
 		public static Texture2D FromStream(GraphicsDevice graphicsDevice, Stream stream)
 		{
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Confirmed this by making a fake Stream and tracing Seek/Read calls from XNA Texture2D.FromStream, thanks!

PR #208, opened 2018-11-04
More iOS fixes

TheSpydog:
Just a few small changes needed for iOS/tvOS. Of particular note is that `MediaPlayer` is a namespace in Xamarin.iOS, so I had to change FrameworkDispatcher to refer to `Media.MediaPlayer`. If there's a nicer way to resolve that namespace/classname conflict, that'd be great.


src/FrameworkDispatcher.cs
```diff
@@ -12,7 +12,6 @@
 
 using Microsoft.Xna.Framework.Audio;
 using Microsoft.Xna.Framework.Input.Touch;
-using Microsoft.Xna.Framework.Media;
```
flibitijibibo: Does changing this to `using MediaPlayer = Microsoft.Xna.Framework.Media.MediaPlayer` fix it?

src/FNAPlatform/OpenGLDevice.cs
```diff
@@ -663,6 +647,17 @@ ref wmInfo
 			MojoShader.MOJOSHADER_glMakeContextCurrent(shaderContext);
 			FNALoggerEXT.LogInfo("MojoShader Profile: " + shaderProfile);
 
+			// UIKit needs special treatment for backbuffer behavior
+			if (SDL.SDL_GetPlatform() == "iOS" || SDL.SDL_GetPlatform() == "tvOS")
```
flibitijibibo: Was the wminfo data inaccurate? If possible I'd like to get the GLuint handles from SysWMInfo instead, but if something's not right with it then we can do this.

flibitijibibo:
Other than the 2 comments this LGTM!

src/FNAPlatform/OpenGLDevice.cs
TheSpydog: Yeah, it always returned 0 for both the Framebuffer and Renderbuffer handles, which just put us back where we started.

src/FrameworkDispatcher.cs
TheSpydog: yup!

src/FNAPlatform/OpenGLDevice.cs
flibitijibibo: I wonder if it’s this part here: https://hg.libsdl.org/SDL/file/80f861112154/src/video/uikit/SDL_uikitwindow.m#l357

Depending on the reason for these results this may be an SDL bug.

src/FNAPlatform/OpenGLDevice.cs
```diff
@@ -4726,4 +4728,4 @@ private void ForceToMainThread(Action action)
 
 		#endregion
 	}
-}
+}
```
flibitijibibo: Xamarin pls

PR #210, opened 2018-11-07
Refactoring of OpenGL devices

rds1983:
Following up our discord discussion about adding ability to create effects in run-time from GLSL code. 
I've started up by performing small refactoring that would keep OpenGL API in one place.
As, right now, OpenGLDevice_GL.cs and ModernGLDevice_GL.cs declare almost similar OpenGL API. Basically ModernGLDevice_GL.cs is OpenGLDevice_GL.cs+some more API.
So I've renamed OpenGLDevice_GL.cs to BaseOpenGLDevice.cs and made OpenGLDevice and ModernOpenGLDevice inherit from BaseOpenGLDevice.
Then I've removed all OpenGL API from ModernGLDevice_GL.cs where IntelliSense has been warning about "hides inherited members".
Also I've made OpenGL API of BaseOpenGLDevice class public(keeping class itself internal), so it could be used inside the FNA.

### DIFF
--- a/FNA.csproj
+++ b/FNA.csproj
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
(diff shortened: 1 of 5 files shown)


flibitijibibo:
This doesn’t really do anything meaningful and couples the two devices together when each device is meant to be its own thing. (Not to mention, if VulkanDevice ever gets added I’d like to remove the ModernGLDevice...)

Finish the feature first, that way we know what actually has to change. This refactor as of now is speculative at best.

PR #216, opened 2018-11-27
Orientation Support

TheSpydog:
This PR adds orientation support for mobile devices. Most of the interesting changes are in SDL2_FNAPlatform. This patch depends on SDL 2.0.9 for its orientation features.

Note that this implementation does not allow for changing supported orientations at runtime (via `GraphicsDeviceManager.SupportedOrientation`). Instead, the game's supported orientations must be specified at build time in Info.plist or AndroidManifest.xml, depending on the platform. It sucks that this doesn't match XNA's exact behavior, but there doesn't seem to be a good way to accurately emulate that across modern mobile operating systems. (Besides, I haven't come across a single game or example that uses that feature...)

I have tested this on my iPhone 6S+ and everything works as expected. @0x0ade If you have some time soon, could you play around with this on Android? Should be pretty easy to test, I think -- just draw something to the screen and see if it rotates and scales appropriately for different orientations, and make sure touches are being mapped to the correct location.

One question: should we keep or ditch the `Obsolete` attribute for `SupportedOrientations`? XNA didn't mark the property as obsolete, but this is a rare case where we are actually dumping support for a feature and it works well as a warning to those who would want to use it.

### DIFF
--- a/src/FNAPlatform/FNAWindow.cs
+++ b/src/FNAPlatform/FNAWindow.cs
@@ -155,6 +155,23 @@ namespace Microsoft.Xna.Framework
 
 		#region Protected GameWindow Methods
 
+		protected internal override void SetSupportedOrientations(DisplayOrientation orientations)
+		{
+			/* XNA on Windows Phone had the ability to change
+			 * the list of supported device orientations at runtime.
+			 * Unfortunately, we can't support that reliably across
+			 * multiple mobile platforms. Therefore this method is
+			 * essentially a no-op.
+			 *
+			 * Instead, you should set your supported orientations
+			 * in Info.plist (iOS) or AndroidManifest.xml (Android).
+			 *
+			 * -caleb
+			 */
+
+			FNALoggerEXT.LogWarn("Setting SupportedOrientations has no effect!");
+		}
+
 		protected override void SetTitle(string title)
 		{
 			FNAPlatform.SetWindowTitle(window, title);
--- a/src/FNAPlatform/FNAWindow.cs
+++ b/src/FNAPlatform/FNAWindow.cs
@@ -52,11 +52,8 @@ namespace Microsoft.Xna.Framework
 
 		public override DisplayOrientation CurrentOrientation
 		{
-			get
-			{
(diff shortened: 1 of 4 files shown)


flibitijibibo:
What we might be able to do is use FNALoggerEXT.LogWarn instead of ObsoleteAttribute, since we may in theory be able to implement it some day, it just happens to be nonviable right now.

Other than that this seems fine, so if Android works then we’re good. I’ll try to do the SDL update tomorrow so this works properly.

src/FNAPlatform/FNAWindow.cs
```diff
@@ -149,15 +146,15 @@ internal void INTERNAL_ScreenDeviceNameChanged()
...
+		internal void INTERNAL_OnOrientationChanged()
+		{
+			OnOrientationChanged();
+		}
+
 		#endregion
 
 		#region Protected GameWindow Methods
 
-		protected internal override void SetSupportedOrientations(DisplayOrientation orientations)
```
flibitijibibo: Just realized that this was here - sadly we have to keep this per the spec:

https://docs.microsoft.com/en-us/previous-versions/windows/xna/ff634564%28v%3dxnagamestudio.40%29

This can be another FNALoggerEXT line, I think.

src/GraphicsDeviceManager.cs
```diff
@@ -87,17 +87,26 @@ public bool SynchronizeWithVerticalRetrace
...
+
 			get
 			{
 				return supportedOrientations;
 			}
 			set
 			{
+				FNALoggerEXT.LogWarn("Setting SupportedOrientations has no effect!");
 				supportedOrientations = value;
-				if (game.Window != null)
```
flibitijibibo: ... so this file can go back to the old way, and the warning can go in the FNAWindow function.

TheSpydog: should I keep the comment where it is or move it to FNAWindow too?

flibitijibibo: Let's move it to FNAWindow - the GraphicsDeviceManager is doing the right thing, it's the FNAPlatform that's at fault in this case.

PR #217, opened 2018-11-30
Fix for orientation distortion

TheSpydog:
In certain scenarios it is possible for the backbuffer to become distorted upon calling `graphics.ApplyChanges()`. This is because inside ApplyChanges, the backbuffer's width and height are forced to be PreferredBackBufferWidth/Height and do not flip to the appropriate orientation. (For example, an app held in portrait orientation might be forced to have a backbuffer size of 800x480 instead of 480x800, so all visuals are stretched and misshapen.)

This PR fixes that issue and adds a new function to FNAPlatform for checking whether the device supports orientation changes. Up for debate is whether to include WinRT in that list of orientation-supporting OS's.


flibitijibibo:
Windows 10 might support this, but I don't think anyone's shipping a Store App any time soon...

PR #230, opened 2019-01-11
Sort indexes only.

kiates:
Instead of sorting large structs in place, only sort the sprite indexes and then copy them out to an alternate buffer to draw.

### DIFF
--- a/src/Graphics/SpriteBatch.cs
+++ b/src/Graphics/SpriteBatch.cs
@@ -10,6 +10,7 @@
 #region Using Statements
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Runtime.InteropServices;
 using System.Text;
 #endregion
@@ -87,6 +88,8 @@ namespace Microsoft.Xna.Framework.Graphics
 
 		// Local data stored before buffering to GPU
 		private VertexPositionColorTexture4[] vertexInfo;
+		private int[] sortedIndexes;
+		private VertexPositionColorTexture4[] sortedVertexInfo;
 		private Texture2D[] textureInfo;
 
 		// Default SpriteBatch Effect
@@ -126,9 +129,9 @@ namespace Microsoft.Xna.Framework.Graphics
 		 */
 		private static readonly byte[] spriteEffectCode = Resources.SpriteEffect;
 		private static readonly short[] indexData = GenerateIndexArray();
-		private static readonly TextureComparer TextureCompare = new TextureComparer();
-		private static readonly BackToFrontComparer BackToFrontCompare = new BackToFrontComparer();
-		private static readonly FrontToBackComparer FrontToBackCompare = new FrontToBackComparer();
+		private readonly TextureByIndexComparer textureByIndexComparer;
+		private readonly BackToFrontByIndexComparer backToFrontByIndexComparer;
+		private readonly FrontToBackByIndexComparer frontToBackByIndexComparer;
 
 		#endregion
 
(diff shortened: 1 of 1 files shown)


kiates:
Yeah, I know, I want to find out too. Messes up that pretty consecutive chunk of memory.

I’m going to put together a contrived example that I can compile for both XNA and FNA, maybe even MonoGame if I get into it, in order to closely record comparison metrics. The allocation could be deferred until a sorting spritebatch begin is called (could result in a GC hicup if that call is mid-game), and the alternate sort could be used after a certain number of sprites (but that would be very platform specific with regard to cache size/specifics). Yuck.

Another idea Is to build an alternate vertex index list on the gpu that takes into account the new primitive orderingor, but that goes deeper and I’m not positive what that would look like or if I’d run into design problems. I’m not a OpenGL programmer yet, or a Direct3D one for that matter.

Still another is to change the vertex info struct to a class and manage it as a pool, then the sort can just be pointer swaps instead of the separate index array. We’d still have to make one final gather copy into contiguous memory before sending.

Whatever way, I know there is a significant difference between performance of XNA vs FNA on windows, profiling led me here, and I expect the issue possible contributed to slower performance on Linux/OSX which is one reason I started looking into it. I may log that as a general issue with this PR being one possible optimization. I’ll be running some more tests in engine as well.

If FNA was at least as performant as XNA on windows I’d be a little closer to stopping shipping the XNA version (and dealing with the XNA redist). I do still need to try that Angle trick too, but the OpenGL seems fast so far from what I can tell on my machines.

Sorry for the brain dump. :)

flibitijibibo:
Two things to compare would be sorted and unsorted, since it may also be the GL slowing us down. If we send more than one batch per frame then it has to either stall or sneak in a staging buffer, both of which can slow things down... MESA_DEBUG on Intel Linux will spit messages about this.

For reference, here’s DXTK’s SpriteBatch, which appears to just use std::sort...

https://github.com/Microsoft/DirectXTK/blob/master/Src/SpriteBatch.cpp

Also here’s my SpriteBatchTest, which I used for performance and accuracy tests:

https://gist.github.com/flibitijibibo/27c486c236301b49b650

flibitijibibo:
Closing since this doesn't seem to be the most optimal way to do this...

I think the main bottleneck here is the vertex array - we're doing everything on the array which means we're touching very large chunks of memory for no particularly good reason. So, what we can probably do instead is this:

1. Add SpriteInfo back, containing the parameters sent to PushSprite. Make an array of MAX_SPRITES size. This will be the largest memory addition.
2. Add a SpriteInfo* array that contains references to all the SpriteInfo entries in the above array, use this to sort. This is the second-largest memory addition, but sorting pointers instead of the entire struct will be MUCH nicer.
3. After sorting is done, write to the vertex array by iterating through the sorted array and calling PushSprite for each entry. This makes it so we're only doing the most intensive work once, and therefore not moving huge structs all over the place.

This is pretty much what DXTK does at a very high level, but of course we can't reuse that code exactly for a multitude of reasons. After the above is written my guess is all the SortModes will be way faster, but we need to be careful not to make Deferred slower. Deferred is meant to be the fastest by a considerable margin and I would happily throw the other sort modes into the fire if it just meant having to sort things on the client side instead.

kiates:
I’m playing with something very similar to what your describing and using a modified version of you test. I’ll open up another PR in the future and/or share the test fixtures/stats. One of the things that’s making this difficult is getting reliable times to compare objectively, among other things.

PR #232, opened 2019-01-18
[DNM] Workaround for MFi BigButton bug

TheSpydog:
This is a workaround for [this SDL2 bug](https://bugzilla.libsdl.org/show_bug.cgi?id=4463) regarding MFi controllers and their wonky press/release behavior on iOS/tvOS. Until SDL gets this sorted out on their end, this works and seems reasonable enough.


flibitijibibo:
We now have a functioning patch for this in SDL, please refer to the SDL report in the OP for the proper fix.

PR #235, opened 2019-01-25
(structs & pointers) Speed up sorting SpriteBatch modes by avoiding excessive copying of vertex information

kiates:
Why implement one when you can implement two at twice the cost.

This PR buffers the information used to generate sprite vertex and texture information in an array of SpriteInfo value types (structs), when sorting modes are used, then sorts it using an accompanying array of pointers to those structs, and finally computes the native compatible vertex information array.

In my latest measurements on a Ryzen 7 1800X CPU with AMD 390X GPU, this is consistently a little bit faster than my version that uses an array SpriteInfo classes.

```
Technique: Fna (Before - Stock FNA)
           Type Average     Min     Max  Median
       Deferred    4039    2300    6643    4472
      Immediate   32314   29476   52610   41043
        Texture   15854   11469   28702   20086
    BackToFront   16420   11487   24060   17774
    FrontToBack   17040   11760   29844   20802
```

```
Technique: Fna (SpriteInfo classes)
           Type Average     Min     Max  Median
       Deferred    4043    2409    6765    4587
      Immediate   31869   30341   54410   42376
        Texture    8581    5229   13705    9467
    BackToFront   10567    7107   18253   12680
    FrontToBack   10873    7161   17237   12199
```

This PR:
```
Technique: Fna (SpriteInfo structs and pointers)
           Type Average     Min     Max  Median
       Deferred    3873    2368    6963    4666
      Immediate   32400   29600   52248   40924
        Texture    7522    4962   11525    8244
    BackToFront    9458    6385   15803   11094
    FrontToBack    9489    6372   16014   11193
```

### DIFF
--- a/src/Graphics/SpriteBatch.cs
+++ b/src/Graphics/SpriteBatch.cs
@@ -1134,25 +1134,28 @@ namespace Microsoft.Xna.Framework.Graphics
 			{
 				fixed (SpriteInfo* spriteInfo = &spriteInfos[numSprites])
 				{
-					spriteInfo->textureHash = texture.GetHashCode();
-					spriteInfo->sourceX = sourceX;
-					spriteInfo->sourceY = sourceY;
-					spriteInfo->sourceW = sourceW;
-					spriteInfo->sourceH = sourceH;
-					spriteInfo->destinationX = destinationX;
-					spriteInfo->destinationY = destinationY;
-					spriteInfo->destinationW = destinationW;
-					spriteInfo->destinationH = destinationH;
-					spriteInfo->color = color;
-					spriteInfo->originX = originX;
-					spriteInfo->originY = originY;
-					spriteInfo->rotationSin = rotationSin;
-					spriteInfo->rotationCos = rotationCos;
-					spriteInfo->depth = depth;
-					spriteInfo->effects = effects;
+					fixed (IntPtr* sortedSpritesPtr = &sortedSprites[0])
+					{
+						spriteInfo->textureHash = texture.GetHashCode();
+						spriteInfo->sourceX = sourceX;
+						spriteInfo->sourceY = sourceY;
+						spriteInfo->sourceW = sourceW;
+						spriteInfo->sourceH = sourceH;
+						spriteInfo->destinationX = destinationX;
+						spriteInfo->destinationY = destinationY;
+						spriteInfo->destinationW = destinationW;
+						spriteInfo->destinationH = destinationH;
+						spriteInfo->color = color;
+						spriteInfo->originX = originX;
+						spriteInfo->originY = originY;
+						spriteInfo->rotationSin = rotationSin;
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Whew, so the sorts were _really_ bad then. Also good is that Immediate isn't any worse, but the Deferred has me a bit antsy. The improvement to the average seems good, but somehow the min, max, _and_ median are all higher...? Any idea what's up with that?

Basically as long as it's not _worse_ I'm down with this version, but I wonder if there's something to the data that I'm missing.

flibitijibibo:
Weird as it sounds, running D3D11 via ANGLE might produce different results too:

https://github.com/FNA-XNA/FNA/wiki/4:-FNA-and-Windows-API#direct3d-support

AMD OpenGL is just kinda slow in general (unless you’re on Mesa Linux), so maybe that will highlight something unrelated to the sprite processing.

src/Graphics/SpriteBatch.cs
```diff
@@ -144,6 +147,8 @@ public SpriteBatch(GraphicsDevice graphicsDevice)
 
 			vertexInfo = new VertexPositionColorTexture4[MAX_SPRITES];
 			textureInfo = new Texture2D[MAX_SPRITES];
+			spriteInfos = new SpriteInfo[MAX_SPRITES];
```
flibitijibibo: Just realized that this would have to be either pinned or AllocHGlobal'd, otherwise the GC could move the location of the array around mid-batch and cause all the sortedSprite pointers to be the wrong location and crash.

flibitijibibo: In this case I was referring to `spriteInfos` - for example:

1. spriteInfos allocated, memory location is `0xC000`
2. spriteBatch.Begin();
3. spriteBatch.Draw(...) called 5 times, pointers are `0xC000` through, say, `0xC040`
4. GC acts, decided to move spriteInfos, memory location is now `0xF000`
5. spriteBatch.Draw(...) called once, pointer is `0xC050`
6. spriteBatch.End() called, sorts sprites
7. Sorter tries to access `0xC000`, crash!

So spriteInfos has to be persistently-pinned, or it has to be allocated off the GC heap entirely.

kiates: Bingo, thanks Ethan. I hate garbage collection.

Rather than go to the point of allocating from a different heap, I moved things closer together so they could be pinned together during the critical transformation.  Maybe it will use the cache better too overall.

Running the same stats (though different day so who knows what my PC is juggling) I get...
```

Technique: Fna
           Type Average     Min     Max  Median
       Deferred    3816    2367    6771    4569
      Immediate   34637   30988   55296   43142
        Texture    7445    5122   11383    8253
    BackToFront    9401    6440   19293   12867
    FrontToBack   10202    6413   16153   11283
```

Looks unchanged performance wise.

PR #238, opened 2019-02-03
Clamp glPixelStorei alignment for texture uploads to 8 as demanded by the spec

kg:
Right now if I try to upload a Vector4 texture, pixelstorei fails because the requested alignment (16 bytes) is too much.

It seems like uploads still work right with the clamping set to 8.

### DIFF
--- a/src/FNAPlatform/OpenGLDevice.cs
+++ b/src/FNAPlatform/OpenGLDevice.cs
@@ -2713,8 +2713,9 @@ namespace Microsoft.Xna.Framework.Graphics
 			}
 			else
 			{
-				// Set pixel alignment to match texel size in bytes
-				int packSize = Texture.GetFormatSize(format);
+				// Set pixel alignment to match texel size in bytes. Maximum allowed value is 8
+                // https://www.khronos.org/registry/OpenGL-Refpages/es2.0/xhtml/glPixelStorei.xml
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Huh! I wonder why this didn’t show up when I was testing all these. 2 things:

1. Add clamp to the other PixelStorei further down, for Cubes
2. More specific citation we can use while we fix that tabbing: OpenGL 2.1 specification, section 3.6.1, table 3.1:

https://www.khronos.org/registry/OpenGL/specs/gl/glspec21.pdf

kg:
It looks like there are actually a bunch of glPixelStorei calls in the codebase, do I need to fix all of them?

flibitijibibo:
Only the ones that set it away from 4, half of them are for keeping the state sane afterward. You’re right though, ModernGLDevice will have this problem too.

src/Graphics/Texture.cs
```diff
@@ -97,6 +97,13 @@ internal static int GetFormatSize(SurfaceFormat format)
 			}
 		}
 
+		internal static int GetPixelStoreAlignment(SurfaceFormat format) {
```
flibitijibibo: Brace on left and /**/, then this is set to go! (You can sign this comment if you want, too)

PR #240, opened 2019-02-25
Fix ShowRuntimeError crash

TheSpydog:
Fixes a crash on .NET Core when showing a runtime error message due to the title or message being null.


flibitijibibo:
Haven't heard back on this and I'm not 100% sure how easy this is to do on every target, so for now we'll just do this ourselves...

PR #242, opened 2019-03-10
Use polling for TouchPanel.GetState + FingerIdEXT

TheSpydog:
This PR contains a few fixes/changes to the Touch namespace. Namely...

1. The kinda wonky event-based touch processing for TouchPanel.GetState has been replaced with a polling system that scans all the fingers currently active on the screen and compares them to their previous state. This is probably closer to what XNA did internally, and this may fix an annoyingly elusive bug where touches weren't registered as released when they should be. (Haven't 100% verified this fixes it, though.) Gesture detection is still event-driven, though!

2. I replaced the multitude of `-1`'s hanging out in GestureDetector and TouchPanel with an actual constant (`TouchPanel.NO_FINGER`). This makes things clearer, I think, and it also makes it easier to change in the future if we find out -1 is a valid finger id on some platform.

3. I added two extension properties to GestureSample: `FingerIdEXT` and `FingerId2EXT`. These give the finger ids involved in a gesture (or `NO_FINGER` if not applicable), which is very useful for storing finger info and connecting registered gestures with TouchLocations. I added this because I had a need to prevent a finger from registering as a press on virtual gamepad buttons after it had sent a certain Gesture event. There is no other reliable way to gather this information, so I thought this might qualify as a good FNA extension.

### DIFF
--- a/src/FNAPlatform/SDL2_FNAPlatform.cs
+++ b/src/FNAPlatform/SDL2_FNAPlatform.cs
@@ -2567,33 +2567,27 @@ namespace Microsoft.Xna.Framework
 			};
 		}
 
-		public static void UpdateTouchPanelState()
+		public static unsafe void UpdateTouchPanelState()
 		{
 			// Poll the touch device for all active fingers
 			long touchDevice = SDL.SDL_GetTouchDevice(0);
 			for (int i = 0; i < TouchPanel.MAX_TOUCHES; i += 1)
 			{
-				IntPtr fingerPtr = SDL.SDL_GetTouchFinger(touchDevice, i);
-				if (fingerPtr == IntPtr.Zero)
+				SDL.SDL_Finger* finger = (SDL.SDL_Finger*) SDL.SDL_GetTouchFinger(touchDevice, i);
+				if (finger == null)
 				{
 					// No finger found at this index
 					TouchPanel.SetFinger(i, TouchPanel.NO_FINGER, Vector2.Zero);
 					continue;
 				}
 
-				// Get finger info
-				SDL.SDL_Finger finger = (SDL.SDL_Finger) Marshal.PtrToStructure(
-					fingerPtr,
-					typeof(SDL.SDL_Finger)
-				);
-
 				// Send the finger data to the TouchPanel
 				TouchPanel.SetFinger(
 					i,
-					(int) finger.id,
+					(int) finger->id,
 					new Vector2(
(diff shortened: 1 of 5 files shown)


src/FNAPlatform/SDL2_FNAPlatform.cs
```diff
@@ -2567,6 +2567,38 @@ public static TouchPanelCapabilities GetTouchCapabilities()
...
+				IntPtr fingerPtr = SDL.SDL_GetTouchFinger(touchDevice, i);
+				if (fingerPtr == IntPtr.Zero)
+				{
+					// No finger found at this index
+					TouchPanel.SetFinger(i, TouchPanel.NO_FINGER, Vector2.Zero);
+					continue;
+				}
+
+				// Get finger info
+				SDL.SDL_Finger finger = (SDL.SDL_Finger) Marshal.PtrToStructure(
```
flibitijibibo: Any chance we can turn this into an `unsafe { SDL.SDL_Finger* finger; }`? Just to avoid the variable duplication and such.

flibitijibibo:
This all seems fine to me, just needs some documentation to go into https://github.com/FNA-XNA/FNA/wiki/5:-FNA-Extensions. Write up as much (or as little) as you think is needed and I'll throw it into the wiki at the same time I merge.

TheSpydog:
How about this for wiki documentation?

**FingerIdEXT / FingerId2EXT**

`public int GestureSample.FingerIdEXT` and `public int GestureSample.FingerId2EXT` are new properties for GestureSample that allow you to determine exactly which touch finger(s) were involved in the gesture. 

Note that `FingerId2EXT` is only used for `Pinch` and `PinchComplete` gestures, and it will have a value of -1 for any other gesture type.

The following is a basic example of using these properties:
```
// Keep track of which fingers were used for these gestures
private int DragFinger = -1;
private int PinchFinger1 = -1;
private int PinchFinger2 = -1;

// Somewhere later...

while (TouchPanel.IsGestureAvailable)
{
    GestureSample gesture = TouchPanel.ReadGesture();
    if (gesture.GestureType == GestureType.Drag)
    {
        DragFinger = gesture.FingerIdEXT;
    }
    else if (gesture.GestureType == GestureType.Pinch)
    {
        PinchFinger1 = gesture.FingerIdEXT;
        PinchFinger2 = gesture.FingerId2EXT;
    }
}
```

PR #245, opened 2019-04-01
[WIP] Implement PipelineCache for Effect state changes

flibitijibibo:
This one's mostly for @TheSpydog, but is important to anyone who cares about explicit APIs showing up in FNA some day.

Right now one of the absolute worst hacks in FNA is the state cache hack in Effect. While the GraphicsDevice is D3D10-like, XNA still uses D3D9 effects and thus still uses D3D9 render state change commands in its render passes. This means that to properly support effect render states, I had to come up with a horrendously bloated technique to generate pipeline objects based on previous pipeline state objects and any changes that the effect pass would produce. This happens to work okay since OpenGL doesn't have PSOs, but all new graphics APIs now have them, so this hack is now a giant roadblock for anyone trying to create a new GLDevice.

This branch will fix this issue with what I'm calling the "PipelineCache", which as of now moves all the Effect cache hack work into a single shared class owned by the GraphicsDevice. Functionally it should be no different other than reducing the number of pipeline objects allocated at once. The next thing that needs to be done is to generate hash maps for each type of object, which means we need to figure out how to generate hash codes for those objects. This is easier said than done considering all the states have large values in them, so don't expect to fit all the state values into a 4-to-8-byte value. This will definitely be the hardest part.

Once that's done we just have to use, for example, a `Dictionary<HashType, BlendState>` to store all the unique objects, and then we can completely remove the cache hack and have a proper cache to store each unique PSO generated by the Effects framework, meaning explicit graphics APIs will have a MUCH more pleasant time managing that branch of memory. (We'll also have to introduce a `IGLPSO` interface in IGLDevice, but that's for another day.)

### DIFF
--- a/FNA.Core.csproj
+++ b/FNA.Core.csproj
@@ -211,6 +211,7 @@
 		<Compile Include="src\Graphics\PackedVector\Rgba1010102.cs" />
 		<Compile Include="src\Graphics\PackedVector\Short2.cs" />
 		<Compile Include="src\Graphics\PackedVector\Short4.cs" />
+		<Compile Include="src\Graphics\PipelineCache.cs" />
 		<Compile Include="src\Graphics\PresentationParameters.cs" />
 		<Compile Include="src\Graphics\PresentInterval.cs" />
 		<Compile Include="src\Graphics\PrimitiveType.cs" />
--- a/FNA.csproj
+++ b/FNA.csproj
@@ -279,6 +279,7 @@
     <Compile Include="src\Graphics\PackedVector\Rgba1010102.cs" />
     <Compile Include="src\Graphics\PackedVector\Short2.cs" />
     <Compile Include="src\Graphics\PackedVector\Short4.cs" />
+    <Compile Include="src\Graphics\PipelineCache.cs" />
     <Compile Include="src\Graphics\PresentationParameters.cs" />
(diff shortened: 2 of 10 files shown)


flibitijibibo:
@TheSpydog has written up all the hashing work! All I did was fold it all into PipelineCache and make the hash functions do the math against the variables in PipelineCache. I think this has legitimate uses not just for Effects but for XNA3->XNA4 ports, so I made it all 1 file to make it convenient for possible reuse in such a port.

Testing for accuracy will be important, and performance will be worth looking at as well - the hash functions are probably okay but the Dictionary lookup for a struct key type might be a bit scary...

flibitijibibo:
We ended up making a 128-bit type to do the hashing, so performance should be a lot better now. I dunno if it'll be as fast as the old stuff, but that'll be where accuracy makes up for it. If anybody has Effects that change render state aggressively we'd love to have them.

TheSpydog:
What are we waiting on for this PR? More example effects for testing?

flibitijibibo:
Fell behind on this, I'll check on the last few examples I have for this and squash it into master if everything works.

flibitijibibo:
It turns out I was using this branch by accident for like 2 weeks, so I think this is good...

PR #253, opened 2019-05-17
Invalidate FBOs at the start of each frame

TheSpydog:


src/FNAPlatform/OpenGLDevice.cs
```diff
@@ -1059,6 +1059,28 @@ out dstH
...
 			}
 
+			/* Invalidate this at the very beginning of the frame.
+			 * This appears to prevent incoherent memory accesses
+			 * on iOS, though we're not sure why it has to be here
+			 * and not at SetRenderTargets like it should be.
+			 */
+			if (supportsFBOInvalidation)
+			{
+				glBindFramebuffer(GLenum.GL_FRAMEBUFFER, realBackbufferFBO);
```
flibitijibibo: Does this still work if you switch targetFramebuffer and realBackbufferFBO? This will make it so we preserve the correct framebuffer binding afterward.

TheSpydog: Apparently not. The order matters here for some reason. I have an idea that might fix that... I'll investigate

PR #256, opened 2019-06-06
Updated Intellisense descriptions

foreverWIP:
The description for DPadUp referred to down, and vice versa for DPadDown. Swapped the two descriptions to correct them


flibitijibibo:
Shows how often some of these docs get looked at... thanks for the fix!

PR #259, opened 2019-07-12
Add missing generic type forwards to XNA ABI-compatibility libraries.

zfigura:
This fixes DLC Quest running under wine-mono, and probably several other games as well.


flibitijibibo:
Interesting, I wonder if we have to worry about other <T>s? In any case, thanks for the fix!

zfigura:
I checked FNA.dll for all public classes, and those were the only two I found that weren't in the abi/*.cs files.

flibitijibibo:
Good enough for me!

PR #262, opened 2019-07-15
Replace Queue<int> for sampler modification tracking with bool[]

kg:
The Queue.Contains call was showing up in my profiles (about 1% of total CPU) and doesn't seem necessary.

I can pull out the comments once you're sure the behavior is right

### DIFF
--- a/src/Graphics/GraphicsDevice.cs
+++ b/src/Graphics/GraphicsDevice.cs
@@ -291,8 +291,9 @@ namespace Microsoft.Xna.Framework.Graphics
 
 		#region Internal Sampler Change Queue
 
-		private readonly Queue<int> modifiedSamplers = new Queue<int>();
-		private readonly Queue<int> modifiedVertexSamplers = new Queue<int>();
+
+		private readonly bool[] modifiedSamplers = new bool[MAX_TEXTURE_SAMPLERS];
+		private readonly bool[] modifiedVertexSamplers = new bool[MAX_VERTEXTEXTURE_SAMPLERS];
 
 		#endregion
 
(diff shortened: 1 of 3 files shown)


src/Graphics/SamplerStateCollection.cs
```diff
@@ -38,20 +35,21 @@ public sealed class SamplerStateCollection
...
 		internal SamplerStateCollection(
 			int slots,
-			Queue<int> modSamplers
+			bool[] modSamplers
 		) {
 			samplers = new SamplerState[slots];
 			modifiedSamplers = modSamplers;
 			for (int i = 0; i < samplers.Length; i += 1)
 			{
+				// FIXME: Shouldn't we be setting modifiedSamplers here?
```
flibitijibibo: Only if the GL default sampler parameters doesn’t match. Thankfully this appears to be consistent for every target, and of course the default init could be defined at IGLDevice initialization.

src/Graphics/GraphicsDevice.cs
```diff
@@ -1467,24 +1468,29 @@ private void ApplyState()
...
 				GLDevice.VerifySampler(
 					sampler,
 					Textures[sampler],
 					SamplerStates[sampler]
 				);
 			}
-			while (modifiedVertexSamplers.Count > 0)
-			{
+
+			for (int sampler = 0; sampler < modifiedVertexSamplers.Length; sampler++) {
```
flibitijibibo: += 1, brace on next line

src/Graphics/GraphicsDevice.cs
```diff
@@ -1467,24 +1468,29 @@ private void ApplyState()
...
+				if (!modifiedVertexSamplers[sampler])
+					continue;
```
flibitijibibo: Braces

src/Graphics/GraphicsDevice.cs
```diff
@@ -1467,24 +1468,29 @@ private void ApplyState()
 				RenderTargetCount > 0
 			);
 
-			while (modifiedSamplers.Count > 0)
-			{
-				int sampler = modifiedSamplers.Dequeue();
+			for (int sampler = 0; sampler < modifiedSamplers.Length; sampler++) {
```
flibitijibibo: Same notes as vertex samplers

src/Graphics/GraphicsDevice.cs
```diff
@@ -291,8 +291,9 @@ internal int RenderTargetCount
 
 		#region Internal Sampler Change Queue
 
-		private readonly Queue<int> modifiedSamplers = new Queue<int>();
-		private readonly Queue<int> modifiedVertexSamplers = new Queue<int>();
+
```
flibitijibibo: Extra space

PR #266, opened 2019-08-12
Use dictionary for SpriteFont index lookups

TheSpydog:
Tested with Rogue Legacy: Wanderer Edition on an iPhone XR. This change reduces the time spent in Draw() overall by ~10%, and the per-DrawString performance impact has been reduced 8x. From my testing there's no significant memory impact... if anything, this actually made the RAM usage slightly smaller. (No idea how/why. May be a fluke.)

This needs testing on desktop platforms, but I feel pretty good about its performance on ARM.


flibitijibibo:
Holy crap, okay then...

We may still want to look at @nkast’s suggestions from the Discord, but for now, this is more than enough for me. Even if x86’s performance wasn’t as good this is more than worth the tradeoff.

PR #273, opened 2019-10-01
FNA.dll.config -> app.config

AntyMew:
MSBuild automatically recognizes `app.config` as a configuration file to be copied to the build output, renamed as `FNA.dll.config`. This is, on paper, the same behavior as the current project.

However, given a .NET Core project referencing FNA via ProjectReference, when attempting to package the project as a single-file executable (`/p:PublishSingleFile=true`), the file gets counted twice: once by auto-detecting `FNA.dll.config` in the build output, once by the explicit `CopyToOutputDirectory`. This trips an [exception](https://github.com/dotnet/core-setup/blob/6c40a810faacb904c623a5d2575b424298f0e37b/src/managed/Microsoft.NET.HostModel/Bundle/Bundler.cs#L163) in the single-file bundler, failing the build.

Renaming `FNA.dll.config` to `app.config` resolves this issue.


flibitijibibo:
This is not an appropriate change, as it breaks compatibility with virtually every other build and runtime configuration in existence.

This either needs to be filed as an msbuild issue or the fix needs to factor in compatibility with... well, anything.

flibitijibibo:
For added context, test the following setups:

- Visual Studio 2010, 2015, 2019
- MonoDevelop 5 with xbuild, 7 with msbuild

flibitijibibo:
It looks like I misunderstood the OP's writeup - it made it sound like it was a .NET Core thing, when in reality it's a part of .NET Framework (the current version goes as far back as 2.0):

https://support.microsoft.com/en-us/help/815786/how-to-store-and-retrieve-custom-information-from-an-application-confi

This works with VS2010 and xbuild, which are the worst of the build systems we support.

Apologies for the mixup, my natural instinct for Core patches is usually to assume the worst, since most Core users don't care about compatibility...

PR #274, opened 2019-10-24
Remove Color(Color, alpha) constructors

0x0ade:
This PR simply removes the following `Color` constructor signatures which don't exist in XNA 4.0:
- `Color(Color color, int alpha)`
- `Color(Color color, float alpha)`

They're [not listed on ~~MSDN~~ Microsoft Docs](https://docs.microsoft.com/en-us/previous-versions/windows/xna/ff433831%28v%3dxnagamestudio.41%29) and are currently causing issues in the Celeste modding community, where mods are built once with either FNA or XNA and then relinked to XNA / FNA depending on which version of the game is running:

```
System.MissingMethodException: Method not found: 'Void Microsoft.Xna.Framework.Color..ctor(Microsoft.Xna.Framework.Color, Single)'.
```

### DIFF
--- a/src/Color.cs
+++ b/src/Color.cs
@@ -1598,40 +1598,6 @@ namespace Microsoft.Xna.Framework
 		}
 
 		/// <summary>
-		/// Constructs an RGBA color from a <see cref="Color"/> and an alpha value.
-		/// </summary>
-		/// <param name="color">
-		/// A <see cref="Color"/> for RGB values of new <see cref="Color"/> instance.
-		/// </param>
-		/// <param name="alpha">The alpha component value from 0 to 255.</param>
-		public Color(Color color, int alpha)
-		{
-			packedValue = 0;
-
-			R = color.R;
-			G = color.G;
-			B = color.B;
-			A = (byte) MathHelper.Clamp(alpha, Byte.MinValue, Byte.MaxValue);
-		}
-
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Huh, I wonder why this doesn’t show up in our [spec test](https://github.com/flibitijibibo/XNA4SpecTest)?

0x0ade:
https://github.com/flibitijibibo/XNA4SpecTest/blob/4010572b10b367ff637621ecd6d8cc5b9c25fca4/Program.cs#L197-L242

It checks fields, properties, events, methods, but it does not check constructors.

flibitijibibo:
I could swear we did that as well, what an obvious thing to miss... we should probably run that before 19.11!

flibitijibibo:
Seems like it’s suddenly treating internals as public... we should probably fix this up and get it under the FNA-XNA organization so it can be maintained properly.

Sineaggi:
Does Bastion depend on this constructor? `Color(Color color, float alpha)` seems like it's still used.
Screen Shot 2020-03-24 at 6 15 51 PM

flibitijibibo:
Yes, as does Flinthook and Mercenary Kings. No, we won’t add it back. Hack it in if it’s mandatory, otherwise request an update from the developer.

PR #278, opened 2019-12-11
Add RefreshRateEXT

TheSpydog:
The refresh rate of DisplayModes may be useful for some games. In particular, Chasm may need this information to prevent an issue with the game screen turning black at some resolutions. To avoid spamming everyone with duplicate display modes, I added the compile define `DISPLAYMODE_KEEP_DUPLICATES`.

Very experimental!

### DIFF
--- a/src/FNAPlatform/SDL2_FNAPlatform.cs
+++ b/src/FNAPlatform/SDL2_FNAPlatform.cs
@@ -7,6 +7,19 @@
  */
 #endregion
 
+#region DISPLAYMODE_KEEP_DUPLICATES Option
+// #define DISPLAYMODE_KEEP_DUPLICATES
+/* By default, FNA will remove all duplicates from the list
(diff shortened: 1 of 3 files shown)


flibitijibibo:
We could do more precise filtering based on the active desktop mode, but this is definitely a red herring. We never modify the display mode _ever_, so the refresh rates advertised are irrelevant since all modes are either a window or a fullscreen window with a faux-backbuffer, which can be entirely arbitrary in size.

The black screen is probably something dumber like an old Intel driver or an overlay program interfering with framebuffer bindings.

TheSpydog:
Makes sense. This type of bug has been reported by a [few](https://steamcommunity.com/app/312200/discussions/0/1753520224838820712/) [people](https://steamcommunity.com/app/312200/discussions/0/1753520125914856656/), so something screwy's going on, but it's probably not this. Closing...

PR #279, opened 2019-12-19
Removed allocations due to the lambda action .ctor

andreakarasho:
At the moment when `DISABLE_THREADING` is defined FNA calls the method `ForceToMainThread` and here a tons of actions are allocated.
Basically when you do:

```csharp
ForceToMainThread(() => { Console.WriteLine("Hello"); });
```
you are explicing allocate a new Action:
```csharp
ForceToMainThread(new Action(() => { Console.WriteLine("Hello"); }));
```

With my PR you will create and push an action when ThreadID is not equals to the main one.


flibitijibibo:
I can’t tell what the actual changes are in this patch because the formatting is a disaster. Please fix this first.

PR #280, opened 2019-12-23
Remove allocations when DISABLE_THREADING is not defined

andreakarasho:
The current FNA code implements this logic structure to push data to the main thread from a different one:

```csharp
public class ADevice : IGLDevice
{
    public void A_GL_MethodCall()
    {
#if !DISABLE_THREADING
        ForceToMainThread() => {
#endif
			/* ...
			/* code
			/* ...
			*/
#if !DISABLE_THREADING
			});
#endif
    }
}
```

This patch aims to remove the implicit ```new Action( () => {} );``` allocation in runtime in this way:

```csharp
public class ADevice : IGLDevice
{
    public void A_GL_MethodCall()
    {
#if !DISABLE_THREADING
	if (mainThreadId == Thread.CurrentThread.ManagedThreadId)
	{
#endif

		/* ...
		/* code
		/* ...
		*/

#if !DISABLE_THREADING
	}
	else 
	{		 
		ForceToMainThread() => {

			/* ...
			/* code
			/* ...
			*/

		});
	}
#endif
    }
}
```

The code formatting should be the same now.

### DIFF
--- a/FNA.Core.csproj
+++ b/FNA.Core.csproj
@@ -129,6 +129,8 @@
 		<Compile Include="src\FNAPlatform\FNAPlatform.cs" />
 		<Compile Include="src\FNAPlatform\FNAWindow.cs" />
 		<Compile Include="src\FNAPlatform\IGLDevice.cs" />
+		<Compile Include="src\FNAPlatform\MetalDevice.cs" />
+		<Compile Include="src\FNAPlatform\MetalDevice_MTL.cs" />
 		<Compile Include="src\FNAPlatform\ModernGLDevice.cs" />
 		<Compile Include="src\FNAPlatform\ModernGLDevice_GL.cs" />
 		<Compile Include="src\FNAPlatform\OpenGLDevice.cs" />
--- a/FNA.csproj
+++ b/FNA.csproj
@@ -197,6 +197,8 @@
     <Compile Include="src\FNAPlatform\FNAPlatform.cs" />
     <Compile Include="src\FNAPlatform\FNAWindow.cs" />
     <Compile Include="src\FNAPlatform\IGLDevice.cs" />
+    <Compile Include="src\FNAPlatform\MetalDevice.cs" />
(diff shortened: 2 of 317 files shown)


flibitijibibo:
This is a lot better!

From what I can tell it should work, but as you probably noticed, this adds a _ton_ of duplicate code, making it very very hard to maintain. Any change to the GL call stream requires making the same one change a minimum of 2 times, and sometimes a maximum of 4 times! That’s a lot of stress for FNA to bear.

Something worth looking at is WineD3D’s Command Stream Multi-threading:

https://github.com/wine-mirror/wine/blob/master/dlls/wined3d/cs.c

A better fix for this problem is a command stream system that offloads all commands to a dedicated thread, which would still have allocations but make threaded GLDevice calls completely safe without blocking for a whole frame. Additionally, we could make all the allocations in the unmanaged heap, since we could malloc the command structs instead of needing to make a `new Action`. This was partially started in the form of ThreadedGLDevice...

https://github.com/FNA-XNA/FNA/blob/master/src/FNAPlatform/ThreadedGLDevice.cs

... but this doesn’t work well with OpenGLDevice due to issues with the state machine and command ordering, and it still uses Actions instead of command structs.

If you absolutely need these Actions gone, consider making ThreadedGLDevice fit your needs while DISABLE_THREADING is enabled in OpenGLDevice. The sooner someone makes ThreadedGLDevice work, the sooner I can make that the default and remove all the threading code completely.

TheSpydog:
This PR is in really rough shape after those merges. Could this be cleaned up so there's not 318 changed files?

flibitijibibo:
Closing this, since we now support build options in a much more convenient manner:

https://github.com/FNA-XNA/FNA/commit/d6c18b2dcf97a4024ca39ddbef07a499f0585461#diff-666ff6e6e1eb2c434463f9938dede9f4

If you use threaded GL and don't funnel the GL calls properly, that's the application's bug, not ours.

PR #281, opened 2019-12-25
Add FNA_PREFER_SDL2_BASE_PATH to prefer SDL's base path

nickgravelyn:
Because I'm pedantic about macOS I wanted to have my content live in the Resources folder of my app bundle but FNA has logic to just use the assembly directory by default. This might be fine for most people so I thought I could add an environment variable around it such that I can skip ahead to just using the SDL2 base path (which by default is the bundle resource directory).

The comment in there makes it sound like eventually that whole chunk will go away but I have no idea the full ramifications there. Likely a lot of breaking of games that didn't seem super necessary for what I wanted to achieve right now.

### DIFF
--- a/src/FNAPlatform/SDL2_FNAPlatform.cs
+++ b/src/FNAPlatform/SDL2_FNAPlatform.cs
@@ -1409,23 +1409,26 @@ namespace Microsoft.Xna.Framework
 
 		private static string GetBaseDirectory()
 		{
-			if (	OSVersion.Equals("Windows") ||
-				OSVersion.Equals("Mac OS X") ||
-				OSVersion.Equals("Linux") ||
-				OSVersion.Equals("FreeBSD") ||
-				OSVersion.Equals("OpenBSD") ||
-				OSVersion.Equals("NetBSD")	)
+			if (Environment.GetEnvironmentVariable("FNA_PREFER_SDL2_BASE_PATH") != "1")
 			{
-				/* This is mostly here for legacy compatibility.
-				 * For most platforms this should be the same as
-				 * SDL_GetBasePath, but some platforms (Apple's)
-				 * will have a separate Resources folder that is
-				 * the "base" directory for applications.
-				 *
-				 * TODO: Remove this and endure the breakage.
-				 * -flibit
-				 */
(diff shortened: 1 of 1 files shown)


flibitijibibo:
The main reason SDL_GetBasePath isn't the default is because of system Mono installations - when using /usr/bin/mono you get /usr/bin as the base rather than the location of the entry exe. Not that you should distribute with that kind of dependency, but for development on Linux and macOS it can get kind of annoying...

Is `FNA_SDL2_FORCE_BASE_PATH` an okay name? Mostly trying to keep the `FNA_$SUBSYSTEM_$DESCRIPTION` namespacing. Otherwise this is fine with me!

src/FNAPlatform/SDL2_FNAPlatform.cs
```diff
@@ -1409,23 +1409,26 @@ public static void SetRelativeMouseMode(bool enable)
...
+			if (Environment.GetEnvironmentVariable("FNA_PREFER_SDL2_BASE_PATH") != "1")
+			{
+				if (	OSVersion.Equals("Windows") ||
+					OSVersion.Equals("Mac OS X") ||
+					OSVersion.Equals("Linux") ||
+					OSVersion.Equals("FreeBSD") ||
+					OSVersion.Equals("OpenBSD") ||
+					OSVersion.Equals("NetBSD")	)
+				{
+					/* This is mostly here for legacy compatibility.
```
flibitijibibo: For some reason the editor didn't put the spaces back into the `*` lines... why are editors like this now?

nickgravelyn: Probably my fault since I normally use 2 spaces for indentation so I had to change up Sublime to tabs and reindent it all. I added the spaces back in.

nickgravelyn:
All makes sense. I changed the environment variable; I have no real opinions on the names so whatever you want works for me.

flibitijibibo:
Perfect, thanks!

flibitijibibo:
Just finished a restructure of MonoKickstart that makes it so this is effectively the default behavior on macOS. (Full announcement is in the Discord)

PR #282, opened 2019-12-26
Fix high DPI for macOS initial window size

nickgravelyn:
So I think this is a good change but feel free to educate me if I'm missing the point on something. The problem I spotted was that I enabled high DPI mode for my game using `FNA_GRAPHICS_ENABLE_HIGHDPI` and in my game's constructor I initialize the GraphicsDeviceManager with a preferred size of 1280x720. When the game starts up my window is 640x360 with a 1280x720 backbuffer. While that is technically what I specified it wasn't what I expected. 

What I found was that I could move this little bit of code around from the resize event and into ApplyChanges and now I get what I would expect which is a 1280x720 window and a 2560x1440 backbuffer. This makes more sense as it ensures that my game initialization code doesn't have to know if the platform supports high DPI; whether it does or does not I still get a 720p window which is what I would hope. I guess part of the issue is that the XNA API requires setting the back buffer resolution rather than the window size.

Regardless of this change obviously the rendering has to deal make sure to read the back buffer size instead of window size to properly handle high DPI modes but that's to be expected and doesn't require any knowledge of high DPI or not; it just means using one width/height versus another.

Maybe there's a reason to not do it this way but it seems more intuitive to me. This is also kind of a breaking change, though, since games already handling high DPI mode are going to be twice as big on startup now. What do you think?

### DIFF
--- a/src/GraphicsDeviceManager.cs
+++ b/src/GraphicsDeviceManager.cs
@@ -337,6 +337,13 @@ namespace Microsoft.Xna.Framework
 					}
 				}
 			}
+
+			if (Environment.GetEnvironmentVariable("FNA_GRAPHICS_ENABLE_HIGHDPI") == "1")
+			{
+				gdi.PresentationParameters.BackBufferWidth *= 2;
+				gdi.PresentationParameters.BackBufferHeight *= 2;
+			}
+
 			gdi.PresentationParameters.DepthStencilFormat =
 				PreferredDepthStencilFormat;
 			gdi.PresentationParameters.IsFullScreen =
@@ -470,11 +477,6 @@ namespace Microsoft.Xna.Framework
 			Rectangle size = (sender as GameWindow).ClientBounds;
 			resizedBackBufferWidth = size.Width;
 			resizedBackBufferHeight = size.Height;
-			if (Environment.GetEnvironmentVariable("FNA_GRAPHICS_ENABLE_HIGHDPI") == "1")
-			{
-				resizedBackBufferWidth *= 2;
-				resizedBackBufferHeight *= 2;
-			}
 			useResizedBackBuffer = true;
 			ApplyChanges();
 		}


flibitijibibo:
When originally writing this feature in I wasn't sure which would be right, but after using it a few times in practice and looking at the spec more closely, the current implementation is the only right answer we have.

In practice, consider a resolution selector. In our current implementation, the small resolutions can be pretty small, but the render size will still be correct, and as you pick the highest resolutions the window will still fit on the desktop. In the case where you double the _window_ size, you'll get as far as the desktop res, then every res on top of that goes past the desktop size, all the way to the native Retina res where it's now 4 times larger than the desktop (and you don't even wanna think about how big that drawable size is)! That's assuming windowed mode, of course, but then you get to fullscreen - even if we doubled _windowed_ mode, the reality is that fullscreen mode would be doing _exactly what our current halving logic does anyway_. This is easy to explain with the spec:

The method that XNA programs use to modify the window size is not actually with the _window_ size, but the _backbuffer_ size. Up until now, they just happened to be the same by coincidence. But now that they're different, we have to choose whether to prioritize the window size or the backbuffer size. In the case of XNA, the choice is surprisingly easy, since programs _explicitly_ specify, for example, `PreferredBackBufferWidth`, and can always query `PresentationParameters.BackBufferWidth` either from the GraphicsDevice or from a Reset event. The window is never mentioned once! This made it easy for me to say "who cares what the window size is, just specify the backbuffer dimensions and the window will accommodate that demand," and then it turned out that SDL_WINDOW_FULLSCREEN_DESKTOP forced that conclusion anyhow.

The only way you would ever know the _window_ size is if you use `Window.ClientBounds` or `OnClientSizeChanged`, both of which were, even in the original XNA spec, platform-specific and easy to manipulate into doing undefined behavior. The annoying thing part is that OnClientSizeChanged isn't even the best way to detect resizes, the best method is actually `DeviceReset`... it always fires on real resizes (as the patch shows, GraphicsDeviceManager already uses OnClientSizeChanged internally), and it ends up returning the information that actually matters in the first place. This is useful particularly for [certain](https://github.com/FNA-XNA/FNA/wiki/Appendix-B:-FNA-on-Consoles#nintendo-switch) [platforms](https://github.com/FNA-XNA/FNA/wiki/Appendix-A:-FNA-on-iOS-and-tvOS) that have resizes, but at the same time don't have "windows" at all...

TL;DR: For high-DPI window support, regardless of platform, the best way to go is to pretend there is no such thing as a window in the first place.

nickgravelyn:
Yeah that makes sense from the framework standpoint. XNA was entirely designed around exclusive full screen on Xbox where there is no window so it doesn't have great APIs for true windowed mode. And high DPI just wasn't really a thing then.

All of what you're saying makes total sense for full screen cases, especially setting the back buffer to a good size for fullscreen windows. Would you just let the startup window be half size for windowed mode? Otherwise I also can do something like this in my `LoadContent` method:

```cs
if (
  !_graphics.IsFullScreen &&
  GraphicsDevice.PresentationParameters.BackBufferWidth > Window.ClientBounds.Width &&
  GraphicsDevice.PresentationParameters.BackBufferHeight > Window.ClientBounds.Height)
{
  _graphics.PreferredBackBufferWidth *= 2;
  _graphics.PreferredBackBufferHeight *= 2;
  _graphics.ApplyChanges();
}
```

It's a bit of a hack to handle one specific case on startup, but it seems to be working for me. Basically I want to solve for the case that my game's window can appear in two different "physical" sizes on the same computer depending on whether the monitor is high DPI or not which feels weird to me as a player. 

Do many games even have windowed mode these days? Am I crazy for even caring this much about it? 🤣

flibitijibibo:
Didn't think about the init case - I suppose we can resize the window if this block trips:

https://github.com/FNA-XNA/FNA/blob/master/src/FNAPlatform/SDL2_FNAPlatform.cs#L497

Does resizing the window in this block fix the issue? If so I can merge that in, provided @TheSpydog doesn't see any issue elsewhere.

nickgravelyn:
That block does trigger on my Mac but I'm not sure if resizing there will do anything, since at that point it's still sized using the default resolution (i.e. my preferred resolution from my game's constructor hasn't been applied). So at some point after that it's likely that my resolution preference will trigger and the window will get shrunk back down.

It probably does make sense to leave it as is for consistency since otherwise you end up in a weird case where sometimes setting back buffer size means window size and other times just back buffer size. It's not terribly hard to handle in my game code anyway and I want to make sure I don't make my life harder than need be for future Xbox, Switch, etc support.

PR #286, opened 2020-01-06
[WIP] New hashing method for PipelineCache

TheSpydog:
I've been using this method for hashing inside MetalDevice for a while now, and I thought it might make sense to bring it over to PipelineCache.

The main gist is that instead of custom 128-bit structs, we use uints for hashing. This speeds up dictionary lookups considerably (can't get much faster than an int!) but introduces a small risk of hash collisions. If somehow two PSOs of the same kind generated the same hash and were stored into a dictionary under that key, then retrieving one of them would result in the implementation picking one of them arbitrarily. Worst case scenario, this could result in improper drawing behavior. Like I said, though, I have yet to see any evidence of a key collision in any of the games I've tested with Metal.

My main motivation with bringing this over to PipelineCache is to get it out of MetalDevice. It seemed strange to duplicate most of the hashing functions for a single GLDevice. I'd rather keep them all in one place.

I've tested this PR with Murder Miners and it works without issue. (Surprisingly few games actually use PipelineCache... I also tried Cryptark and Capsized but didn't see any VERBOSE_PIPELINECACHE output.)

### DIFF
--- a/src/Graphics/PipelineCache.cs
+++ b/src/Graphics/PipelineCache.cs
@@ -17,63 +17,11 @@
 #endregion
 
 #region Using Statements
-using System;
 using System.Collections.Generic;
-using System.Runtime.InteropServices;
 #endregion
 
 namespace Microsoft.Xna.Framework.Graphics
 {
-	#region Internal PSO Hash Struct
-
-	[StructLayout(LayoutKind.Sequential, Pack = 4, Size = 128)]
-	internal struct StateHash : IEquatable<StateHash>
-	{
-		readonly int i1;
-		readonly int i2;
-		readonly int i3;
-		readonly int i4;
-
-		public StateHash(int i1, int i2, int i3, int i4)
-		{
-			this.i1 = i1;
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Codewise lgtm. We should push every possible state through this and see what collisions come out, if any. There’s probably a legit way to compress this down to a uint but currently it would be impossible to diagnose and fix if you didn’t know this was a thing. Of course, we can also do a long, still one var and probably just as fast.

TheSpydog:
After more brute-force testing, it seems this method yields a substantial number of collisions. I tried switching to `ulong` and that seemed to work pretty okay. At least, I didn't see any collisions before the program ran out of memory and crashed...

The possibility space is unbelievably enormous. We'd need a few petabytes of RAM to accurately check against all possible collisions. Not super feasible for me and my MacBook.

I'm going to close this since I think I've found a better way to optimize the existing system in such a way that actually allows for verifiable collision-proofing. I'll submit a separate PR for that.

PR #287, opened 2020-01-06
remove SO version suffix from OpenBSD to avoid clash with OS library …

rfht:
…cranks

We crank library versions pretty aggressively on OpenBSD and libFAudio is already out of sync. Better to leave out the SO version suffix on OpenBSD and leave it to the porter (me).

### DIFF
--- a/app.config
+++ b/app.config
@@ -3,16 +3,17 @@
 	<dllmap dll="SDL2" os="windows" target="SDL2.dll"/>
 	<dllmap dll="SDL2" os="osx" target="libSDL2-2.0.0.dylib"/>
 	<dllmap dll="SDL2" os="linux,freebsd,netbsd" target="libSDL2-2.0.so.0"/>
-	<dllmap dll="SDL2" os="openbsd" target="libSDL2.so.0"/>
+	<dllmap dll="SDL2" os="openbsd" target="libSDL2.so"/>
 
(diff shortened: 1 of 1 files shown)


flibitijibibo:
I believe those names are automatically searched by the runtime, does it still work if you remove those lines entirely?

rfht:
well, the fnaify system on OpenBSD loads a system-installed mono config, so removing the whole file fixes things. The only problem is that target `libFAudio.so` in there gets overridden by `libFAudio.so.0` in the `FNA.dll.config` if it exists, and the installed library version doesn't match that anymore.

It's otherwise getting questionable if openbsd shoud be in app.config at all.

flibitijibibo:
I’ll leave it up to you - my preference is to remove it if the so versioning is too fragile; normally we try to be specific to prevent compatibility issues (for example, assuming libSDL2.so like Mono does will be really bad when SDL 2.1 breaks ABI), but if the minor version isn’t usable either then I would throw it out and track it in the mono config instead.

flibitijibibo:
Ended up resolving this by just removing the OpenBSD entries:

https://github.com/FNA-XNA/FNA/commit/686d85f2ebd7dc0e824cb143d531d5a29ed58e49

The explicit entries aren't necessary since Mono already does this (though I still don't like that unversioned names are the best thing that it can come up with).

PR #288, opened 2020-01-06
PipelineCache Optimization

TheSpydog:
This update to PipelineCache contains two optimizations:
1. No longer specifying Pack and Size on StateHash. This substantially increased performance on my MacBook under Mono.
2. StateHash now uses two `ulong` values rather than four ints. Less variables to compare = speedup.

Benchmarked the changes against the old version with [this test program](https://gist.github.com/TheSpydog/a645a12527da655ade7c8f0721ff179b). Compared to the previous version, this is 5-6x faster with Mono and 3x faster with .NET Core 3.0, all while retaining the collision-proof hashing we all know and love.

### DIFF
--- a/src/Graphics/PipelineCache.cs
+++ b/src/Graphics/PipelineCache.cs
@@ -29,10 +29,10 @@ namespace Microsoft.Xna.Framework.Graphics
 	[StructLayout(LayoutKind.Sequential)]
 	internal struct StateHash : IEquatable<StateHash>
 	{
-		readonly long a;
-		readonly long b;
+		readonly ulong a;
+		readonly ulong b;
 
-		public StateHash(long a, long b)
+		public StateHash(ulong a, ulong b)
 		{
 			this.a = a;
 			this.b = b;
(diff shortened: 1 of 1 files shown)


src/Graphics/PipelineCache.cs
```diff
@@ -681,9 +672,9 @@ public void EndApplySampler(SamplerStateCollection samplers, int register)
 
 		#region Private Helper Methods
 
-		private static unsafe int FloatToInt32(float f)
+		private static unsafe ulong FloatToUInt64(float f)
 		{
-			return *((int *) &f);
```
flibitijibibo: This will probably have to stay as a uint conversion because of the type sizes, followed by a ulong cast at the end.

Aside from that I think all the signed<->unsigned conversions need the `unchecked` keyword to prevent exceptions (and maybe optimize things further?).

TheSpydog: Good catch, fixed.

flibitijibibo:
Ran a couple type casting tests locally, we're probably going to have to do this to be 100% sure the conversion is stable:

```
int example = 1;
unchecked
{
    ulong max = (ulong) example << 63;
    System.Console.WriteLine(max);
}
```

Basically: ulongs everywhere, with `unchecked` to ensure no meddling or Exceptions from the runtime.

PR #293, opened 2020-02-24
Add support for configuring FNA project externally by placing a props file in the solution directory

kg:
This changes FNA.csproj so that if '$(SolutionDir)FNA.Settings.props' exists, it will be included after other project properties are set. The user can use this to (among other things) set DISABLE_THREADING or THREADED_GL without having to modify the csproj.

### DIFF
--- a/FNA.csproj
+++ b/FNA.csproj
@@ -10,7 +10,8 @@
     <RootNamespace>Microsoft.Xna.Framework</RootNamespace>
     <AssemblyName>FNA</AssemblyName>
   </PropertyGroup>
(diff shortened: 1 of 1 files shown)


flibitijibibo:
20.03 ended up being super quiet so I'm up for getting this in within the next day or so. I don't want 48ac516 but I'd be up for applying 6a96abb if we can get the example props file written. If this works for FNA.Core as well we can include it there too.

flibitijibibo:
Worked on this with Cosmonaut's help, thanks for coming up with this idea!

https://github.com/FNA-XNA/FNA/commit/f5bc93bb4f71fdc164717279857dc3b68a03f6f7

PR #300, opened 2020-03-26
FNA3D SetRenderTargets and ResolveTarget

TheSpydog:


src/Graphics/FNA3D.cs
```diff
@@ -370,15 +386,16 @@ int vertexOffset
 		[DllImport(nativeLibName, CallingConvention = CallingConvention.Cdecl)]
 		public static extern void FNA3D_SetRenderTargets(
 			IntPtr device,
-			/* FIXME: Oh shit RenderTargetBinding[] renderTargets, */
+			IntPtr renderTargets, /* FNA3D_RenderTargetBinding[] */
```
flibitijibibo: I'm surprised arrays don't work here... the GCHandles seem a bit like overkill, does using `fixed (FNA3D.FNA3D_RenderTargetBinding* bindings = &nativeTargetBindings[0])` work if arrays by themselves don't?

flibitijibibo: Disregard arrays, I suck, the marshaler kills performance if you do that. Try using `fixed` instead of GCHandle and see if it still works after that.

flibitijibibo:
`git am`'d this patch and did the fixed keyword work, should be all set now!

PR #301, opened 2020-03-27
Replace IGLDevice with FNA3D

flibitijibibo:
This is the issue tracker for replacing our graphics subsystem with a C99 library called FNA3D:

https://github.com/FNA-XNA/FNA3D/

This will be merged after FNA 20.08 releases.


src/Graphics/GraphicsDevice.cs
```diff
@@ -1565,6 +1549,102 @@ private unsafe void PrepareVertexBindingArray(int baseVertex)
...
+			{
+				if (userIndexBuffer != IntPtr.Zero)
+				{
+					FNA3D.FNA3D_AddDisposeIndexBuffer(
+						GLDevice,
+						userIndexBuffer
+					);
+				}
+
+				/* Initialize the buffer with 2x the needed size.
```
TheSpydog: Is this actually initializing the buffer with 2x the size? If not that's fine, just probably oughta remove the comment. :)

flibitijibibo: Ended up cutting it in the latest revision - at first it was a direct copypaste of the d3d11 renderer, and discovered that it was comparing the non-multiplied length anyway. We can stress check this at some point just to be sure.

PR #311, opened 2020-05-18
Eliminate comparison garbage by providing an explicit comparer for IntPtr

kiates:
Value type and dictionaries don't get along, but creating a dedicated Comparer will keep it from boxing.

### DIFF
--- a/src/Audio/AudioEngine.cs
+++ b/src/Audio/AudioEngine.cs
@@ -65,8 +65,23 @@ namespace Microsoft.Xna.Framework.Audio
 		private readonly FAudio.FACTNotificationCallback xactNotificationFunc;
 		private FAudio.FACTNotificationDescription notificationDesc;
 
+		private class IntPtrComparer : IEqualityComparer<IntPtr>
+		{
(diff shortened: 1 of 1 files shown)


src/Audio/AudioEngine.cs
```diff
@@ -65,8 +65,23 @@ public bool IsDisposed
...
+		private class IntPtrComparer : IEqualityComparer<IntPtr>
+		{
+			public bool Equals(IntPtr x, IntPtr y)
+			{
+				return x == y;
+			}
+
+			public int GetHashCode(IntPtr obj)
+			{
+				return (int)obj;
```
flibitijibibo, line 77: This may need to be ToInt32()... but at the same time, the pointer truncation makes me a bit uncomfortable. Is it even used? If not, we can just do ToInt32 and call it a day, otherwise we may need to use obj.GetHashCode(), goofy as that sounds.

kiates: Best to delegate to its gethash code.

Honestly, in this particular case, I don't think it uses it.

flibitijibibo:
Works for me, thanks!

PR #314, opened 2020-05-29
[WIP] NativeLibrary fallback for DllMap

TheSpydog:
Despite repeated attempts to convince the .NET Foundation to include DllMap support in CoreCLR, it doesn't seem like that's going to happen. Apparently even Mono might be disabling this feature by default now, so we need to be prepared for this.

This PR lays the groundwork for using the NativeLibrary API when running with .NET Core 3.0+. The goal is to have native library importing work _exactly_ like DllMap, by automatically consuming the `FNA.dll.config` / `app.config` file and mapping the library names accordingly. There should be no difference from a user's perspective.

The implementation is mostly based on [this gist](https://gist.github.com/clarvalon/95dfc426838cb60d95dea1abc558d824) written by @clarvalon (used with permission!) but modified to be more appropriate for direct inclusion in FNA and to avoid any licensing issues with some of the third-party code included in the gist.

A few things to note:
1. All the new code is contained inside `src/Utilities/DllMap.cs`, with one internal setup function that gets called in Game's constructor. Changes to existing code consist of exactly 1 extra line.
2. All of the NativeLibrary-relevant code has been gated off by `#if NETCOREAPP3_0` compile flags, so if you're using Mono or .NET Framework, none of this will affect you.
3. In addition to the NativeLibrary changes, this class also folds [x86/x64 Windows search path handling](https://github.com/FNA-XNA/FNA/wiki/4:-FNA-and-Windows-API#64-bit-support) into the DllMap class. I think this makes sense, since pretty much every FNA project should have this anyway for good behavior on Windows, but if there are good reasons to remove it and require this on the application side, I'm totally fine with that.
4. When .NET 5 officially comes out, we may want to change the `NETCOREAPP3_0` target flags into `NET5_0` or whatever the equivalent is. I'm not sure how backward-compatibility works with target framework flags.

I've tested this on macOS Catalina and it appears to work indistinguishably from mono. However, this will need **very thorough** testing on all supported platforms to make sure this works as expected!


flibitijibibo:
We’ll want to remove the reference to System.Xml.Linq, as that’s always a garbage festival. I would also like to find a way to statically initialize this, so it runs as soon as FNA is executed at all - having it in Game isn’t the best place for this, but FNAPlatform’s static constructor would be better if we can do a standalone static init.

flibitijibibo:
Also, we will probably want to mark this as netstandard2.1, rather than Core specifically, since numerous targets using other runtimes will be affected by this.

TheSpydog:
Closing this in favor of #315.

PR #315, opened 2020-05-29
Dynamically load CoreDllMap assembly

TheSpydog:
This code detects if the NativeLibrary type exists in the current runtime, and if so, tries to dynamically load the [CoreDllMap](https://github.com/TheSpydog/CoreDllMap) assembly in the same directory. All the DllMap reimplementation logic is contained inside the other DLL, so it removes the burden from FNA.

Still needs testing with .NET Framework, UWP, and possibly BRUTE(?) to make sure the changes didn't screw up anything.

### DIFF
--- a/FNA.Core.csproj
+++ b/FNA.Core.csproj
@@ -13,6 +13,9 @@
 	<PropertyGroup>
 		<FNASettingsPropsFilePath>$(SolutionDir)FNA.Settings.props</FNASettingsPropsFilePath>
 	</PropertyGroup>
+	<ItemGroup>
+		<PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0" PrivateAssets="All" />
+	</ItemGroup>
 	<Import Project="$(FNASettingsPropsFilePath)" Condition="Exists('$(FNASettingsPropsFilePath)')" />
 	<Target Name="ValidatePropsFilePath" BeforeTargets="BeforeBuild">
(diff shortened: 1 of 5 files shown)


TheSpydog:
Tested with iOS and it works without issue. No extra dependencies are pulled as far as I can tell.

TheSpydog:
From testing it appears that there are many games that call into SDL2-CS before any calls are made in the XNA namespace. This is problematic, since we don't have a chance to register the DllMap before the call is made. As I see it, there are two possible solutions to this:
1. Add static constructors to SDL2-CS, FAudio, MojoShader, and Theorafile. (FNA3D is likely unaffected since it's only called internally.) We can basically copy-paste the CoreDllMap init code into each of these constructors to be 100% sure it's ready to roll by the time PInvokes start getting called. _However..._ this incurs a performance overhead. How much overhead is hard to say without profiling, and different runtimes will likely be affected to different degrees. If needed, there are ways to mitigate the performance losses, but they might get ugly.
2. Kick this responsibility entirely over to the user side. If they want to make a game with .NET Core, they'll need to call `CoreDllMap.Register` themselves, at the top of their Main function. I have mixed feelings about this idea. It totally lifts the burden of DllMap initialization off our shoulders, but at the same time, it also removes the ability to run existing games with .NET Core (without source modification, at least). Despite the quirks involved, I like being able to run `dotnet Flotilla.exe`. It'd be cool to make sure games are somewhat future-proofed by automatically giving them the capability to run under CoreCLR. (Excluding incompatibilities out of our hands, of course.)

Thoughts?

TheSpydog:
After more consideration, both options suck. Option 1 sacrifices performance for non-Core builds for no good reason, and Option 2 is literally exactly where we started with this whole mess. Closing.

Issue #316, opened 2020-06-30
DynamicSoundEffect references in FrameworkDispatcher.Update() are sometimes null.


ZaneDubya:
Running 20.06.

My bug reporter has collected a small number of null reference exceptions from players. The null reference exception is thrown from [FrameworkDispatcher.cs at line 40](https://github.com/FNA-XNA/FNA/blob/3d470062728d4879fbbb47ab45f853bf870f1a5c/src/FrameworkDispatcher.cs#L40).

I instance and dispose of DynamicSoundEffect instances on a second thread, but I do not directly modify the FrameworkDispatcher.Streams collection from my code.

I edited the linked Update() routine as follows, and have not seen any crashes since. My changes are:
- Lock Streams because the Streams object can be updated while this method is running,
- Make sure the instance is not null before updating.

I am happy to submit this (in your coding style) as a PR.

        public static void Update() {
            /* Updates the status of various framework components
            * (such as power state and media), and raises related events.
            */
            for (int i = 0; i < Streams.Count; i += 1) {
                DynamicSoundEffectInstance dsfi = null;
                lock (Streams) {
                    if (Streams.Count > i) {
                        dsfi = Streams[i];
                    }
                    if (dsfi == null) {
                        continue;
                    }
                }
                dsfi.Update();
                if (dsfi.IsDisposed) {
                    i -= 1;
                }
            }
            . . .

flibitijibibo:
It seems like the lock should be for the whole loop, not for each access. That should prevent the entries from ever being null.

ZaneDubya:
Yes, that would simplify it:

            lock (Streams) {
                for (int i = 0; i < Streams.Count; i += 1) {
                    DynamicSoundEffectInstance dsfi = Streams[i];
                    dsfi.Update();
                    if (dsfi.IsDisposed) {
                        i -= 1;
                    }
                }
            }

flibitijibibo:
Makes sense to me! If this works for you, send the patch with tabs + left braces and I'll merge it in today.

One of the peculiar behaviors of the DSFI list is that when an effect is Disposed, it is removed from the list, so if we're in this block and it ends with the instance being disposed, we know it's been removed from the list and we need to decrement to avoid skipping over a buffer.

ZaneDubya:
I've opened a PR with tabs and left braces at #317 - let me know if any changes are needed.

Thanks for all your work on this project!

flibitijibibo:
Fixed by #317!

PR #317, opened 2020-06-30
Lock FrameworkDispatcher.Streams while iterating

ZaneDubya:
Addresses possible null reference when a DynamicSoundEffect instance is disposed in another thread while FrameworkDispatcher.Update() is running. See issue #316

### DIFF
--- a/src/Audio/DynamicSoundEffectInstance.cs
+++ b/src/Audio/DynamicSoundEffectInstance.cs
@@ -134,7 +134,10 @@ namespace Microsoft.Xna.Framework.Audio
 
 			// Okay we're good
 			base.Play();
-			FrameworkDispatcher.Streams.Add(this);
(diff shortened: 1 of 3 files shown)


flibitijibibo:
Style looks good, just need the same lock on these two and this is good to go:

https://github.com/FNA-XNA/FNA/blob/master/src/Audio/DynamicSoundEffectInstance.cs#L137
https://github.com/FNA-XNA/FNA/blob/master/src/Audio/SoundEffectInstance.cs#L405

ZaneDubya:
I've added locks to the two references to FrameworkDispatcher.Streams you indicated. I'm happy to mash these into a single commit and open a new PR if you would prefer!

flibitijibibo:
Squashed em on my end, thanks for the fix!

PR #319, opened 2020-07-18
Remove iOS dllmap configs

TheSpydog:
We'll also need to update the iOS wiki page with a section about how when compiling for iOS, you should (1) replace the existing *.dylib entries with __Internal in the config file and (2) add __Internal config entries for libmojoshader and libtheorafile as well.


flibitijibibo:
Send the diff for the wiki and I'll get it in ASAP!

Issue #322, opened 2020-08-02
Split up FNAPlatform.RunLoop


flibitijibibo:
This one's mostly for @TheSpydog who's currently exploring wasm, but also applies to @jfmajor who needs this feature for their editor.

Currently the loop is held entirely in the FNAPlatform function, but this is a problem for 2 reasons:

1. RunOneFrame needs to poll events before Tick(), to handle certain updates (key up/down, hotplugging, text input, things like that)
2. Every FNAPlatform conceivable is going to have a loop anyway, and a lot of code at the top and bottom of RunLoop could easily be shared

This is a proposed refactor, but could change depending on performance/accuracy issues we've not encountered yet:

- [ ] `FNAPlatform.RegisterGame` shows the game window and adds it to `activeGames`
- [ ] `FNAPlatform.UnregisterGame` removes the game from `activeGames`
- [ ] `FNAPlatform.PollEvents` contains the SDL_Event loop, add this to RunOneFrame()
- [ ] `OSXUseSpaces` becomes a static readonly variable
- [ ] `List<Keys> keys` is initialized in Keyboard.cs instead
- [ ] Text input variables become Game instance variables
- [ ] `displayIndex` becomes a `GraphicsAdapter` rather than an `int`
- [ ] Game would get a new `RunLoop` that looks like this:

```
private void RunLoop()
{
    while (RunApplication)
    {
        FNAPlatform.PollEvents(
            this,
            currentAdapter,
            textInputControlDown,
            textInputControlRepeat,
            ref textInputSuppress
        );
        Tick();
    }
    Exit();
}

protected virtual void BeginRun()
{
    if (loopBegan)
    {
        return;
    }

    currentAdapter = FNAPlatform.RegisterGame(this);

    Rectangle windowBounds = Window.ClientBounds;
    Mouse.INTERNAL_WindowWidth = windowBounds.Width;
    Mouse.INTERNAL_WindowHeight = windowBounds.Height;

    // Perform initial check for a touch device
    TouchPanel.TouchDeviceExists = GetTouchCapabilities().IsConnected;

    loopBegan = true;
}

protected virtual void EndRun()
{
    if (loopBegan)
    {
        FNAPlatform.UnregisterGame(this);
        loopBegan = false;
    }
}
```

flibitijibibo:
Fixed by #324

PR #324, opened 2020-08-03
[WIP] Split up FNAPlatform.RunLoop

TheSpydog:
First draft of a fix for #322. Pretty sure TextInputCharacters/Bindings could be static (and maybe readonly?) since they're effectively constant.

I haven't tested text input, but keyboard and gamepad input work as expected. Could definitely use more testing on this!

### DIFF
--- a/src/FNAPlatform/SDL2_FNAPlatform.cs
+++ b/src/FNAPlatform/SDL2_FNAPlatform.cs
@@ -686,25 +686,17 @@ namespace Microsoft.Xna.Framework
 
 		#region Event Loop
 
-		public static void RunLoop(Game game)
+		static bool osxUseSpaces;
+
+		public static GraphicsAdapter RegisterGame(Game game)
 		{
 			SDL.SDL_ShowWindow(game.Window.Handle);
 			game.IsActive = true;
-
-			Rectangle windowBounds = game.Window.ClientBounds;
-			Mouse.INTERNAL_WindowWidth = windowBounds.Width;
-			Mouse.INTERNAL_WindowHeight = windowBounds.Height;
-
-			// Which display did we end up on?
-			int displayIndex = SDL.SDL_GetWindowDisplayIndex(
-				game.Window.Handle
-			);
 
 			// Store this for internal event filter work
 			activeGames.Add(game);
 
 			// OSX has some fancy fullscreen features, let's use them!
-			bool osxUseSpaces;
 			if (OSVersion.Equals("Mac OS X"))
 			{
 				string hint = SDL.SDL_GetHint(SDL.SDL_HINT_VIDEO_MAC_FULLSCREEN_SPACES);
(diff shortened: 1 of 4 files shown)


src/FNAPlatform/SDL2_FNAPlatform.cs
```diff
@@ -941,25 +941,17 @@ public static bool SupportsOrientationChanges()
 
 		#region Event Loop
 
-		public static void RunLoop(Game game)
+		static bool osxUseSpaces;
```
flibitijibibo, line 944: This can be put at the top next to the other static constants, probably as

```
private static readonly bool OSXUseSpaces = (
    SDL.SDL_GetPlatform().Equals("Mac OS X") && // Prevents race with OSVersion
    SDL.SDL_GetHintBoolean(SDL.SDL_HINT_VIDEO_MAC_FULLSCREEN_SPACES, 1) == SDL.SDL_bool.SDL_TRUE
);
```

src/Game.cs
```diff
@@ -183,6 +183,30 @@ public GameWindow Window
 
 		internal bool RunApplication;
 
+		/* Setup Text Input Control Character Arrays
```
flibitijibibo, line 186: Hard to say what's the best spot for this, but my gut says FNAPlatform. They can definitely be `static readonly` no matter where it ends up though.

src/Game.cs
```diff
@@ -400,13 +439,29 @@ public void Run()
...
 
-			FNAPlatform.RunLoop(this);
+			RunLoop();
 
 			EndRun();
 
 			OnExiting(this, EventArgs.Empty);
 		}
 
+		private void RunLoop()
```
flibitijibibo, line 449: Functionally okay, should be put in the `#region Private Methods` section for consistency.

flibitijibibo:
Possible accuracy issue - it turns out RunOneFrame() does not call BeginRun/EndRun at all! It also doesn't seem like those functions are the correct place to do the register/unregister as deliberately overriding those without calling base.*Run() doesn't cause any breakages. So we'll most likely have to turn those implementations into standalone functions and use those instead.

TheSpydog:
> turn those implementations into standalone functions and use those instead

I'm a bit confused, what specifically are you referring to by "those implementations"?

flibitijibibo:
The bodies of BeginRun/EndRun will have to be new private functions instead, leaving BeginRun and EndRun empty again.

flibitijibibo:
Committed the BeginRun/EndRun part of RunOneFrame separately:

https://github.com/FNA-XNA/FNA/commit/23fa2b5e30ff4a1dca28ff78cb9f6b486dfd7eb9

Also found a minor timing fix:

https://github.com/FNA-XNA/FNA/commit/8750327539206bade7e2f7105e5318dde3a4fde7

More info on BeginRun/EndRun: The best guess I can make is that our new `BeforeLoop` should happen after BeginRun, and our new `AfterLoop` should be after EndRun. Other interesting tidbits:

- If you never let it get to the main loop, the window is never shown - you can RunOneFrame forever before that point and it'll always run hidden.
- There seem to be a ton of private variables that track various parts of the framework's lifetime - variables like `inRun` and `endRunRequired` are the interesting ones. What those do and why, I have no idea.

flibitijibibo:
Applied manually with the BeforeLoop/AfterLoop fix:

https://github.com/FNA-XNA/FNA/commit/9b146e45c83fdcdec9896219db9a377bf4991209

PR #325, opened 2020-08-07
Handle Emscripten main loop

TheSpydog:
Implemented a couple new FNAPlatform functions for detecting when a platform-specific main loop is needed (as per discussion in the Discord), and added the necessary callbacks for hooking into the emscripten main loop. It's pretty straightforward, but there are a couple things I'm not sure about:

1. Where does the emscripten-specific code belong? It's not SDL-specific and it feels kinda weird to have it all stuffed inside SDL2_FNAPlatform. I was thinking about making a static EmscriptenHelper class and stuffing the DllImports, delegate, and static `emscriptenGame` variable in there. But then that might be getting too "platform layer"-y to belong outside of an FNAPlatform...

2. The current game exit code basically just cancels the loop and freezes all execution. Audio stops, input stops, and the last image drawn stays permanently drawn to the canvas. This seems kind of awkward but there's not really a better way to handle Game.Exit calls in a browser context. Should we keep this behavior around, or remove it entirely and tell people to not call Exit in their web builds?

### DIFF
--- a/src/FNAPlatform/SDL2_FNAPlatform.cs
+++ b/src/FNAPlatform/SDL2_FNAPlatform.cs
@@ -990,6 +990,56 @@ namespace Microsoft.Xna.Framework
 			}
 		}
 
+		public static bool NeedsPlatformMainLoop()
+		{
+			return SDL.SDL_GetPlatform().Equals("Emscripten");
+		}
+
+		public static void RunPlatformMainLoop(Game game)
+		{
+			if (SDL.SDL_GetPlatform().Equals("Emscripten"))
+			{
+				emscriptenGame = game;
+				emscripten_set_main_loop(
+					RunEmscriptenMainLoop,
+					0,
+					1
+				);
+			}
+		}
+
+		#endregion
+
+		#region Emscripten Main Loop
+
+		// FIXME: Where do these belong?
+
+		private static Game emscriptenGame;
+		private delegate void em_callback_func();
+
+		[DllImport("emscripten", CallingConvention = CallingConvention.Cdecl)]
+		private static extern void emscripten_set_main_loop(
+			em_callback_func func,
+			int fps,
+			int simulate_infinite_loop
+		);
+
+		[DllImport("emscripten", CallingConvention = CallingConvention.Cdecl)]
+		private static extern void emscripten_cancel_main_loop();
+
+		[ObjCRuntime.MonoPInvokeCallback(typeof(em_callback_func))]
+		private static void RunEmscriptenMainLoop()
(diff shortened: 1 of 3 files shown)


src/FNAPlatform/SDL2_FNAPlatform.cs
```diff
@@ -990,6 +990,56 @@ public static void UnregisterGame(Game game)
...
+		[ObjCRuntime.MonoPInvokeCallback(typeof(em_callback_func))]
+		private static void RunEmscriptenMainLoop()
+		{
+			emscriptenGame.RunOneFrame();
+
+			// FIXME: Is this even needed...?
+			if (!emscriptenGame.RunApplication)
+			{
+				emscripten_cancel_main_loop();
+				emscriptenGame.Exit();
```
flibitijibibo, line 1039: This gets called at the end of RunLoop, so only the emscipten call is needed.

src/FNAPlatform/SDL2_FNAPlatform.cs
```diff
@@ -990,6 +990,56 @@ public static void UnregisterGame(Game game)
...
+		{
+			if (SDL.SDL_GetPlatform().Equals("Emscripten"))
+			{
+				emscriptenGame = game;
+				emscripten_set_main_loop(
+					RunEmscriptenMainLoop,
+					0,
+					1
+				);
+			}
```
flibitijibibo, line 1008: May need a NotSupportedException for the else case here.

src/FNAPlatform/SDL2_FNAPlatform.cs
```diff
@@ -990,6 +990,56 @@ public static void UnregisterGame(Game game)
...
+		}
+
+		#endregion
+
+		#region Emscripten Main Loop
+
+		// FIXME: Where do these belong?
```
flibitijibibo, line 1015: This seems fine to me - if someone honestly wants to try an SDL-less Emscripten platform this will be the least of their problems.

flibitijibibo:
Scratch the ProgramExit idea, clearly I don't know what I'm doing... maybe we'll have to document a "please Clear() and Present() before exiting your loop" and the cancel loop call can stay where it is.

src/FNAPlatform/SDL2_FNAPlatform.cs
TheSpydog: Interestingly it continues control flow after the cancel call, so it does hit the Exit. I'm fine with swapping them around though, shouldn't matter too much either way.

PR #329, opened 2020-09-26
Expose SDL_TextEditingEvent

kg:
Because SDL currently disables the IME's composition and candidates windows, it's up to the application to render both of them. The candidates window data is completely hidden so there's nothing we can do about that, but surfacing the composition data from this event allows rendering it correctly, at least for the Japanese IME in my testing.

### DIFF
--- a/src/FNAPlatform/SDL2_FNAPlatform.cs
+++ b/src/FNAPlatform/SDL2_FNAPlatform.cs
@@ -961,16 +961,8 @@ namespace Microsoft.Xna.Framework
 					// Based on the SDL2# LPUtf8StrMarshaler
 					unsafe
 					{
-						byte* endPtr = evt.text.text;
-						if (*endPtr != 0)
-						{
-							int bytes = 0;
-							while (*endPtr != 0)
-							{
-								endPtr++;
-								bytes += 1;
-							}
-
+						var bytes = MeasureStringLength(evt.text.text);
+						if (bytes > 0) {
 							/* UTF8 will never encode more characters
 							 * than bytes in a string, so bytes is a
 							 * suitable upper estimate of size needed
@@ -991,6 +983,26 @@ namespace Microsoft.Xna.Framework
 					}
 				}
 
+				else if (evt.type == SDL.SDL_EventType.SDL_TEXTEDITING) 
+				{
+					unsafe {
+						var bytes = MeasureStringLength(evt.edit.text);
+						if (bytes > 0) {
+							char* charsBuffer = stackalloc char[bytes];
+							int chars = Encoding.UTF8.GetChars(
+								evt.edit.text,
+								bytes,
+								charsBuffer,
+								bytes
+							);
+							var text = new string(charsBuffer, 0, chars);
+							TextInputEXT.OnTextEditing(text, evt.edit.start, evt.edit.length);
+						} else {
+							TextInputEXT.OnTextEditing(null, 0, 0);
+						}
+					}
+				}
+
 				// Quit
 				else if (evt.type == SDL.SDL_EventType.SDL_QUIT)
 				{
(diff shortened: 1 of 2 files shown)


src/FNAPlatform/SDL2_FNAPlatform.cs
```diff
@@ -961,16 +961,8 @@ public static void UnregisterGame(Game game)
...
-						if (*endPtr != 0)
-						{
-							int bytes = 0;
-							while (*endPtr != 0)
-							{
-								endPtr++;
-								bytes += 1;
-							}
-
+						var bytes = MeasureStringLength(evt.text.text);
```
flibitijibibo, line 964: Explicit type instead of var

src/FNAPlatform/SDL2_FNAPlatform.cs
```diff
@@ -991,6 +983,26 @@ public static void UnregisterGame(Game game)
 					}
 				}
 
+				else if (evt.type == SDL.SDL_EventType.SDL_TEXTEDITING) 
+				{
+					unsafe {
+						var bytes = MeasureStringLength(evt.edit.text);
```
flibitijibibo, line 989: Explicit type instead of var

src/FNAPlatform/SDL2_FNAPlatform.cs
```diff
@@ -1008,6 +1020,23 @@ public static void UnregisterGame(Game game)
 			}
 		}
 
+		private unsafe static int MeasureStringLength (byte* ptr) {
+			// Based on the SDL2# LPUtf8StrMarshaler
+			unsafe {
```
flibitijibibo, line 1025: Redundant unsafe

src/Input/TextInputEXT.cs
```diff
@@ -13,6 +13,8 @@
 
 namespace Microsoft.Xna.Framework.Input
 {
+	public delegate void TextEditingEventHandlerEXT (string text, int start, int length);
```
flibitijibibo, line 16: Can this be replaced with Action<string, int, int>? We try to expose as few types as possible, so if not this may have to go inside TextInputEXT to keep everything inside the static class.

flibitijibibo (changes requested):
Aside from that, braces on the left.

flibitijibibo:
Last call for the above changes, would like to have this in 20.10 if possible.

src/FNAPlatform/SDL2_FNAPlatform.cs
```diff
@@ -991,6 +984,30 @@ public static void UnregisterGame(Game game)
...
+						if (bytes > 0) 
+						{
+							char* charsBuffer = stackalloc char[bytes];
+							int chars = Encoding.UTF8.GetChars(
+								evt.edit.text,
+								bytes,
+								charsBuffer,
+								bytes
+							);
+							var text = new string(charsBuffer, 0, chars);
```
flibitijibibo, line 1001:
```suggestion
							string text = new string(charsBuffer, 0, chars);
```

src/FNAPlatform/SDL2_FNAPlatform.cs
```diff
@@ -1008,6 +1025,23 @@ public static void UnregisterGame(Game game)
...
+				{
+					endPtr++;
+					bytes += 1;
+				}
+				return bytes;
+			}
+
+			return 0;
+		}
+
```
flibitijibibo, line 1044:
```suggestion
		private unsafe static int MeasureStringLength(byte* ptr)
		{
			int bytes;
			for (bytes = 0; *ptr != 0; ptr += 1, bytes += 1);
			return bytes;
		}
```

flibitijibibo (changes requested):
Patches attached.

src/FNAPlatform/SDL2_FNAPlatform.cs
```diff
@@ -1008,6 +1025,12 @@ public static void UnregisterGame(Game game)
 			}
 		}
 
+		private unsafe static int MeasureStringLength(byte* ptr)
+		{
+			int bytes;
+			for (bytes = 0; *ptr != 0; ptr += 1, bytes += 1);
+			return bytes;
+		}
```
flibitijibibo, line 1033: Agh dangit, this one's my fault.
```suggestion
		}

```

PR #330, opened 2020-10-04
Draft: Adding support for DDS CubeMaps

YTN0:
Tested this on all the DDS skybox textures I have, and works without issues. I am assuming this would also work for MipMapped CubeMaps, but I don't think those are common (and I don't have any, and wouldn't know how to test those...).

### DIFF
--- a/src/Content/ContentManager.cs
+++ b/src/Content/ContentManager.cs
@@ -282,7 +282,11 @@ namespace Microsoft.Xna.Framework.Content
 				modifiedAssetName = MonoGame.Utilities.FileHelpers.NormalizeFilePathSeparators(
 					Path.Combine(RootDirectoryFullPath, assetName)
 				);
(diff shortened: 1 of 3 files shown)


flibitijibibo:
There are some style issues that need fixing, but I think I'll apply these changes manually - I'll try to compress this a bit, but it does seem to make sense to have them as separate functions rather than one method in Texture.cs.

YTN0:
Sounds good!

Yeah, I considered overloading the DDS function to support both Texture2D and TextureCube, but it started getting messy. Seems better to keep them separate.

I am curious about the style issues though... Since I tried to keep it consistent with everything else 🌝

flibitijibibo:
Manually applied in a separate patch: https://github.com/FNA-XNA/FNA/commit/ce67cfc92d134779dad00596691c2b9e7f9fe9c5

PR #332, opened 2020-10-05
Rename `GLDevice` -> `NativeDevice`

deccer:
`GraphicsDevice` comes with an `internal IntPtr GLDevice` and that kept me thinking whether this is a relic from the past or exists for a reason. I hereby forward the commotion to have `GLDevice` renamed to `NativeDevice` as the latter feels more natural and `GL` keeps me thinking about OpenGL, while said device could be anything from OpenGL, to Vulkan over some Metal- or DX related device.


flibitijibibo:
The name is somewhat historical, but I don't know if it's worth changing unless something critical happens with the original name. Something that relates it to FNA3D might be good too, but this isn't too important IMO.

PR #334, opened 2020-10-22
Fix for acccess violation when audio devices fails to construct a master voice

bartwe:
When failing to create the master voice on a sounddevice, make sure t…he AudioContext is not used and cleaned up correctly

### DIFF
--- a/src/Audio/SoundEffect.cs
+++ b/src/Audio/SoundEffect.cs
@@ -490,7 +490,8 @@ namespace Microsoft.Xna.Framework.Audio
 					IntPtr.Zero
 				) != 0) {
 					FAudio.FAudio_Release(ctx);
-					FNALoggerEXT.LogError(
+                    Handle = IntPtr.Zero;
+                    FNALoggerEXT.LogError(
 						"Failed to create mastering voice!"
 					);
 					return;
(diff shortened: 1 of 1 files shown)


src/Audio/SoundEffect.cs
```diff
@@ -490,7 +490,8 @@ out DeviceDetails
 					IntPtr.Zero
 				) != 0) {
 					FAudio.FAudio_Release(ctx);
-					FNALoggerEXT.LogError(
+                    Handle = IntPtr.Zero;
+                    FNALoggerEXT.LogError(
```
flibitijibibo, line 494:
```suggestion
					Handle = IntPtr.Zero;
					FNALoggerEXT.LogError(
```

src/Audio/SoundEffect.cs
```diff
@@ -647,7 +654,15 @@ out devices
 					return;
 				}
 
-				Context = new FAudioContext(ctx, devices);
+				var context = new FAudioContext(ctx, devices);
+
+				if (context.Handle == IntPtr.Zero) {
+					/* Soundcard failed to configure, bail! */
```
flibitijibibo, line 660:
```suggestion
				FAudioContext context = new FAudioContext(ctx, devices);

				if (context.Handle == IntPtr.Zero)
				{
					/* Soundcard failed to configure, bail! */
```

flibitijibibo (changes requested):
Just style stuff...

PR #335, opened 2020-10-30
Fix for SoundEffect not respecting sampler chunk start/end loop data

FrozenChameleon:
SoundEffect in XNA respects wav sampler chunk start/end data for looping (which can be set with a program such as Wavosaur). Currently FNA SoundEffect does not.

### DIFF
--- a/src/Audio/SoundEffect.cs
+++ b/src/Audio/SoundEffect.cs
@@ -424,7 +424,7 @@ namespace Microsoft.Xna.Framework.Audio
 				while (reader.PeekChar() != -1)
 				{
 					string chunkID = new string(reader.ReadChars(4));
-					uint chunkDataSize = reader.ReadUInt32();
+					int chunkDataSize = reader.ReadInt32();
 					if (chunkID == "smpl") // Sampler Chunk Found
 					{
 						reader.ReadUInt32(); // Manufacturer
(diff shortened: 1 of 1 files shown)


src/Audio/SoundEffect.cs
```diff
@@ -416,6 +419,51 @@ public static SoundEffect FromStream(Stream stream)
...
+						reader.ReadUInt32(); // Product
+						reader.ReadUInt32(); // Sample Period
+						reader.ReadUInt32(); // MIDI Unity Note
+						reader.ReadUInt32(); // MIDI Pitch Fraction
+						reader.ReadUInt32(); // SMPTE Format
+						reader.ReadUInt32(); // SMPTE Offset
+						uint numSampleLoops = reader.ReadUInt32();
+						uint samplerData = reader.ReadUInt32();
+
+						for (int i = 0; i < numSampleLoops; i++)
```
flibitijibibo, line 440:
```suggestion
						for (int i = 0; i < numSampleLoops; i += 1)
```

src/Audio/SoundEffect.cs
```diff
@@ -416,6 +419,51 @@ public static SoundEffect FromStream(Stream stream)
...
+							if (i == 0) // Grab loopStart and loopEnd from first sample loop
+							{
+								samplerLoopStart = start;
+								samplerLoopEnd = end;
+							}
+						}
+
+						if (samplerData != 0) // Read Sampler Data if it exists
+						{
+							reader.ReadBytes((int)samplerData);
```
flibitijibibo, line 458:
```suggestion
							reader.ReadBytes((int) samplerData);
```

src/Audio/SoundEffect.cs
```diff
@@ -416,6 +419,51 @@ public static SoundEffect FromStream(Stream stream)
...
+						}
+					}
+					else // Read unwanted chunk data and try again
+					{
+						reader.ReadBytes((int)chunkDataSize);
```
flibitijibibo, line 463:
```suggestion
						reader.ReadBytes((int) chunkDataSize);
```

flibitijibibo (changes requested):
Just some extremely pedantic style changes, otherwise this is good to go!

PR #342, opened 2021-01-10
Update FNA.csproj

deccer:
Add condition to load app.config
VS2019 (at least 16.8.3) is unable to load app.config with the <Link> element

Version 4.8.3761.0 provided in the condition needs to be verified for vs2010 builds

### DIFF
--- a/FNA.csproj
+++ b/FNA.csproj
@@ -408,7 +408,7 @@
     <Compile Include="lib\Theorafile\csharp\Theorafile.cs" />
   </ItemGroup>
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Looks like the build version is actually `4.6.1055.0`.

Also, I was able to reproduce the skip on VS2010 for AnyCPU - it apparently dislikes not having a PlatformTarget:

https://github.com/FNA-XNA/FNA/commit/4b5d4a9cdcc43374c51f43daec69ffbd8b6ff2ab

deccer (commented):
Ok massaged the confirmed version into the PR.

PR #344, opened 2021-01-23
Fix D3D11 warnings by calling ApplyState more frequently

kg:
In a few of my test cases, the last thing I do is blit my UI render target to the screen, and the first thing I do is set it as my RT and paint to it. This is valid (and I clear the sampler states explicitly before even setting the RTs) but because we only apply state on draw calls, the device state is invalid at the point where FNA sets the render targets.

I added ApplyState calls to SetRenderTargets and Clear because I believe we need to flush them there. There might be other spots too but none occurred to me.

### DIFF
--- a/src/Graphics/GraphicsDevice.cs
+++ b/src/Graphics/GraphicsDevice.cs
@@ -729,7 +729,6 @@ namespace Microsoft.Xna.Framework.Graphics
 
 		public void Clear(ClearOptions options, Vector4 color, float depth, int stencil)
 		{
-			ApplyState();
 			DepthFormat dsFormat;
 			if (renderTargetCount == 0)
 			{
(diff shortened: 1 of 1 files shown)


flibitijibibo:
The device state should be getting fixed up in Clear, as we do in OpenGL:

https://github.com/FNA-XNA/FNA3D/blob/master/src/FNA3D_Driver_OpenGL.c#L1428

This sounds like an implementation-specific issue, one that's unique to sampler objects in D3D11. I would look at that renderer first before blowing apart the entire graphics device.

kg:
Clear is too late, I tested it. It needs to happen before setting render targets. Should I make the change in FNA3D, then?

flibitijibibo:
Yes, if the issue only occurs on D3D11.

src/Graphics/GraphicsDevice.cs
```diff
@@ -852,6 +852,12 @@ public void SetRenderTarget(RenderTargetCube renderTarget, CubeMapFace cubeMapFa
 
 		public void SetRenderTargets(params RenderTargetBinding[] renderTargets)
 		{
+			// D3D11 requires our sampler state to be valid (i.e. not point to any of our new RTs)
+			//  before we call SetRenderTargets. At this point FNA3D does not have a current copy
+            //  of the managed sampler state, so we need to apply our current state now instead of
+            //  before our next Clear or Draw operation.
```
flibitijibibo, line 858:
```suggestion
			//  of the managed sampler state, so we need to apply our current state now instead of
			//  before our next Clear or Draw operation.
```

PR #345, opened 2021-02-04
Mass Refactoring.

Hawkeye4040:
Mass refactoring and code cleanup.


flibitijibibo:
Hilarious.

Hawkeye4040:
What do you mean?

flibitijibibo:
Wait... was this actually a real patch? I thought it was just a fork that accidentally opened to the wrong remote.

Hawkeye4040:
Yeah it's a real patch.

Hawkeye4040:
Take a look I cleaned up a lot.

Hawkeye4040:
Oh I see what you mean I just forked it for now because I was thinking about making some other changes that after looking into it I may not have to make now. Still thinking on that one. But anyway this is a good one to merge to master here.

flibitijibibo:
A single commit that _grossly_ violates style, obliterates spacing, and breaks compatibility with our compiler with no functional change, a patch does not make. This is functionally worthless. Please do not send patches like this to anyone for any software, it's a waste of reviewers' time.

Hawkeye4040:
Okay I guess I won't be helping you anymore since you're all butt hurt that someone cleaned up your sloppy code. Fork it is. You realize you don't own XNA right? It's also not a waste of time. You also contradict yourself quite a bit there. If there was no functional change then how would it break compatibility with the compiler? Violates style? I intentionally didn't use that pass since I could tell our formatting styles were different.

Without being able to hear the stress in your voice I would dare say you're butt hurt.

flibitijibibo:
I can't say I was thrilled with the diff, but the replies were absolutely worth it. Thank you for your time, please do not delete these posts!

PR #349, opened 2021-03-17
Basic FUNCTIONAL XMA2 support

0x0ade:
See https://github.com/FNA-XNA/FAudio/pull/247 for more information and the submodule update.

### DIFF
--- a/src/Audio/SoundEffect.cs
+++ b/src/Audio/SoundEffect.cs
@@ -269,7 +269,6 @@ namespace Microsoft.Xna.Framework.Audio
 			}
 			else if (wFormatTag == 0x166)
 			{
-				// Don't let anyone see this or my head will end up on the chopping board... -ade
 				FAudio.FAudioXMA2WaveFormatEx* xma2 = (FAudio.FAudioXMA2WaveFormatEx*) formatPtr;
 				// dwSamplesEncoded / nChannels / (wBitsPerSample / 8) doesn't always (if ever?) match up.
 				handle.PlayLength = xma2->dwPlayLength;
--- a/src/Audio/SoundEffect.cs
+++ b/src/Audio/SoundEffect.cs
@@ -27,7 +27,7 @@ namespace Microsoft.Xna.Framework.Audio
 			{
 				return TimeSpan.FromSeconds(
 					(double) handle.PlayLength /
-					(double) format.nSamplesPerSec
+					(double) sampleRate
 				);
 			}
 		}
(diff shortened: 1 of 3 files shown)


src/Audio/SoundEffect.cs
```diff
@@ -197,18 +202,37 @@ int loopLength
...
 			/* Buffer format */
-			format = new FAudio.FAudioWaveFormatEx();
-			format.wFormatTag = wFormatTag;
-			format.nChannels = nChannels;
-			format.nSamplesPerSec = nSamplesPerSec;
-			format.nAvgBytesPerSec = nAvgBytesPerSec;
-			format.nBlockAlign = nBlockAlign;
-			format.wBitsPerSample = wBitsPerSample;
-			format.cbSize = 0; /* May be needed for ADPCM? */
+			if (extraData == null) {
```
flibitijibibo, line 211: Braces

src/Content/ContentReaders/SoundEffectReader.cs
```diff
@@ -53,10 +53,38 @@ SoundEffect existingInstance
 			uint nAvgBytesPerSec = Swap(se, input.ReadUInt32());
 			ushort nBlockAlign = Swap(se, input.ReadUInt16());
 			ushort wBitsPerSample = Swap(se, input.ReadUInt16());
-			/* ushort cbSize =*/ input.ReadUInt16();
-
-			// Seek past the rest of this crap (cannot seek though!)
-			input.ReadBytes((int) (formatLength - 18));
+			ushort cbSize = Swap(se, input.ReadUInt16());
+
+			byte[] extra = null;
```
flibitijibibo, line 58: Do we just want to use a byte* and stackalloc to prevent GC activity?

flibitijibibo (changes requested):
This actually looks really good! Just a style thing and a possible memory optimization.

src/Content/ContentReaders/SoundEffectReader.cs
0x0ade: Would be nice, but it'd require memcpy in the SoundEffect constructor to copy from extraData to formatPtr + offs. Buffer.BlockCopy doesn't exist in old .NET runtimes and SDL_memcpy exists but is off limits as this is outside of SDL2_FNAPlatform.

Thinking about it a bit more after talking on it on Discord: maaaaybe only for FNA.Core / a compile time flag? Might not be worth the effort tho 😅

flibitijibibo (approved):
Yeah, the stackalloc's not worth it now that I think about it. This is good to go pending Andrew's review for FAudio.

PR #351, opened 2021-03-23
Fix dllimport for WASM

andreakarasho:


TheSpydog:
This isn't really correct since SDL2 doesn't contain these functions. The reason it happens to work is because mono-wasm statically links all those libraries, so whenever you import something from "SDL2" it's really pulling it from the output binary itself, not the particular library you specify. (For instance, this would work just as well if you put "FNA3D" or "FAudio" for the DllImport name.)

The reason we have it as "emscripten" currently is because back in the old mono/mono days you could pass `--pinvoke-libs emscripten` to packager.exe to register a new valid DllImport name. But that's apparently not a thing anymore since the dotnet/runtime migration.

If you're using Uno for making wasm builds, there's [an official DllImport name](https://github.com/unoplatform/Uno.Wasm.Bootstrap#invoking-emscripten-and-mononet-5-native-functions) for Emscripten functions. Maybe we should use that instead, since this seems to be the recommended (only?) workflow for AOT .NET wasm compilation these days.

andreakarasho:
That's odd because neither works

TheSpydog:
Are you using the latest Uno Framework? How are you linking SDL2 into the build?

andreakarasho:
"__Native" is the solution. 
Thanks @TheSpydog

TheSpydog (approved):

PR #354, opened 2021-06-23
Update FNA.sln to allow default build as Any CPU and not skip build

### LINKED ISSUE #353: FNA.csproj is always skipped when building with VS2019
For some reason the project is skipped in the build, even on a clean rebuild. I haven't had the chance to look into why this happens yet.

deccer:
Turns out there are typos in FNA.sln

"AnyCPU" as a configuration name (the values on the right hand side of the equal sign) is not valid, its "Any CPU"

Tested on VS2019 (latest 16.10), VS2022 Preview 1 (latest 17.0)

Fixes #353

### DIFF
--- a/FNA.sln
+++ b/FNA.sln
@@ -9,8 +9,8 @@ Global
 		Release|x86 = Release|x86
 		Debug|x64 = Debug|x64
 		Release|x64 = Release|x64
-		Debug|AnyCPU = Debug|AnyCPU
-		Release|AnyCPU = Release|AnyCPU
+		Debug|AnyCPU = Debug|Any CPU
+		Release|AnyCPU = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(ProjectConfigurationPlatforms) = postSolution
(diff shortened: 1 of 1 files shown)


flibitijibibo (approved):
Will check on 2010 and merge if all is well... this `Platform` string really was a goof wasn't it.

FNA.sln
```diff
@@ -9,8 +9,8 @@ Global
 		Release|x86 = Release|x86
 		Debug|x64 = Debug|x64
 		Release|x64 = Release|x64
-		Debug|AnyCPU = Debug|AnyCPU
-		Release|AnyCPU = Release|AnyCPU
+		Debug|AnyCPU = Debug|Any CPU
+		Release|AnyCPU = Release|Any CPU
```
flibitijibibo, line 13: Does it still work after making this change?

```suggestion
		Debug|Any CPU = Debug|Any CPU
		Release|Any CPU = Release|Any CPU
```

deccer: Let me check

deccer: It does work too, indeed.

PR #355, opened 2021-07-08
FNADllMap for .NET 5+

TheSpydog:
This is a first draft of a custom DllMap implementation using .NET 5 / C# 9's new module initializer feature and the NativeLibrary API.

There is a single new file that contains all the relevant code. It mimics Mono's dllmap by reading an XML file, parsing the elements, and remapping the dynamic library names accordingly. The implementation is (roughly) based on @clarvalon's dllmap implementation originally written for .NET Core 3, and I think it's quite straightforward.

The real kicker with this PR though is that it introduces a new .csproj, which targets net50. This makes three projects in total: one for Framework/Mono, one for .NET Standard (UWP, iOS), and one for .NET 5+. This is... not ideal. I know Xamarin is an officially supported workload in net6, so I think ultimately we'll be fine on that front, but there's been no word from Microsoft on whether UWP will eventually support .NET 5+.

I couldn't think of a good name for the new .NET 5 project, so I decided to make it "FNA.Core.csproj". I've renamed the old FNA.Core.csproj to FNA.NetStandard.csproj to more accurately reflect its use. Not a fan of this change, but I couldn't think of a better name for the new project. Open to suggestions.

I've tested on Windows and macOS, but I haven't tested Linux yet.


kg:
Could we use an elaborate set of msbuild conditionals to unify .Core and .NetStandard?

FNA.Core.csproj
```diff
@@ -1,6 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
 	<PropertyGroup>
-		<TargetFrameworks>net40;netstandard2.0</TargetFrameworks>
+		<TargetFrameworks>net50</TargetFrameworks>
```
flibitijibibo, line 3: Are we able to keep all three of these in a single csproj? I thought the newer compilers defined stuff like NET5_0 by default.

src/Utilities/FNADllMap.cs
```diff
@@ -0,0 +1,152 @@
...
+				// Ignore entries for other CPUs
+				XmlAttribute cpuAttribute = node.Attributes["cpu"];
+				if (cpuAttribute != null && !cpuAttribute.Value.Contains(cpu))
+				{
+					continue;
+				}
+
+				// Find a mapping
+				string oldLib = node.Attributes["dll"].Value;
+				string newLib = node.Attributes["target"].Value;
```
flibitijibibo, line 129: There's also a particular override for entry point as well, I may have used it in Transistor? Maybe an FMOD game?

flibitijibibo (commented):
Functionally this seems like a good start. It'd be cool to make this as small as humanly possible but UWP may throw us for a loop here if net5 doesn't show up there at some point.

TheSpydog:
After some discussion in the discord, I figured out it's possible to use the existing FNA.Core project to target both netstandard2.0 and net50. However, in order to make this work, this PR will have to remove the net40 target from FNA.Core.csproj. The new minimum .NET Framework version compatible with the FNA.Core project is 4.6.1, as per the .NET Standard 2.0 spec. VS 2022 and OmniSharp are forcing our hand here, since Intellisense has a _very_ bad time when referencing projects that target both Framework and Core.

CC @danielcrenna since you're the one who added the net40 target to FNA.Core.csproj originally, let us know if this will cause any problems for you.

danielcrenna:
No impact to me, I use FNA.Package that adds the DLL map dynamically and have been on net5 for awhile now with no issues.

Thanks for the head’s up, though!

src/Utilities/FNADllMap.cs
TheSpydog: It doesn't look like there's an API for that feature. This is one of the problems @kg identified in [this thread](https://github.com/dotnet/runtime/issues/37213) but I don't think it ever got addressed.

flibitijibibo: Well, shoot... it may be worth checking for all the features that dllmap supports and warning if they're present. Having a conformant parser might help pitch the features needed to be feature complete.

src/Utilities/FNADllMap.cs
```diff
@@ -0,0 +1,152 @@
...
+
+		[ModuleInitializer]
+		public static void Init()
+		{
+			// Get the platform and architecture
+			string os = GetPlatformName();
+			string cpu = RuntimeInformation.ProcessArchitecture.ToString().ToLowerInvariant();
+
+			// Locate the config XML
+			Assembly assembly = Assembly.GetCallingAssembly();
```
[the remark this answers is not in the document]
flibitijibibo: That would violate the original dllmap spec - this may be a case where behavior will diverge between JIT and AOT. Does AOT simply return null or does it throw an Exception?

src/Utilities/FNADllMap.cs
```diff
@@ -0,0 +1,152 @@
...
+			}
+
+			// Parse the XML into a mapping dictionary
+			XmlDocument xmlDoc = new XmlDocument();
+			xmlDoc.Load(xmlPath);
+
+			foreach (XmlNode node in xmlDoc.GetElementsByTagName("dllmap"))
+			{
+				// Ignore entries for other OSs
+				if (!node.Attributes["os"].Value.Contains(os))
```
flibitijibibo, line 115: This is technically optional - if it's not present, it is assumed to be for the currently running OS.

src/Utilities/FNADllMap.cs
```diff
@@ -0,0 +1,152 @@
...
+				{
+					continue;
+				}
+
+				// Ignore entries for other CPUs
+				XmlAttribute cpuAttribute = node.Attributes["cpu"];
```
flibitijibibo, line 121: In addition to `cpu` there is also the annoyingly-named `wordsize`, which actually maps to IntPtr.Size. `cpu` generally gets priority, but `wordsize` is enough to differentiate 32-bit and 64-bit while ignore Intel vs ARM, for example.

flibitijibibo (changes requested):
Mostly just covering the [dllmap specification](https://www.mono-project.com/docs/advanced/pinvoke/dllmap/) as thoroughly as possible. Broad notes:

- `dllentry` is also something to consider, basically just adds a nest layer to what we already have here.
- For all entry attributes, there is also the `!` operator, which ignores any value listed in the string.

The only attrib we wouldn't be covering at this point is `name`, which I'm okay with just having a warning for.

src/Utilities/FNADllMap.cs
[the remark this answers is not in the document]
lithiumtoast: You can also use `AppContext.BaseDirectory`.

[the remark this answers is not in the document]
andreakarasho: Few attemps compiled with bflat:

## Assembly.GetCallingAssembly()
```csharp
using System;

var asm = System.Reflection.Assembly.GetCallingAssembly();
			
if (asm == null)
{
	Console.WriteLine("asm is null");
}
else
{
	Console.WriteLine(asm);
}
```

produces the following result:

```
Unhandled Exception: System.PlatformNotSupportedException: Operation is not supported on this platform.
   at System.Reflection.Assembly.GetCallingAssembly() + 0x5e
   at Test.Program.Main(String[]) + 0x9
   at Test!<BaseAddress>+0x2ae97a
```

---


## Environment.CurrentDirectory

```csharp
using System;


Console.WriteLine("Environment.CurrentDirectory = {0}", Environment.CurrentDirectory);
```


produces:

```
Environment.CurrentDirectory = C:\<my_folder_path>\Test
```


---

## AppContext.BaseDirectory

[thanks @lithiumtoast i do not aware about this]

```csharp
using System;


Console.WriteLine("AppContext.BaseDirectory = {0}", AppContext.BaseDirectory);
```


produces [notice the final backslash]:

```
AppContext.BaseDirectory = C:\<my_folder_path>\Test\ 
```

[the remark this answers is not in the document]
flibitijibibo: Very well - we'll need to try GetCallingAssembly first, then catch and fall back to AppContext.BaseDirectory for safety.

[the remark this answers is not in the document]
TheSpydog: Hmm, so how are you supposed to use NativeLibrary with AOT? [SetDllImportResolver](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.nativelibrary.setdllimportresolver) takes an assembly parameter.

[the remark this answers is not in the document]
TheSpydog: It looks like Assembly.GetExecutingAssembly might work...? https://github.com/dotnet/runtimelab/blob/feature/NativeAOT/src/tests/Interop/NativeLibrary/Callback/CallbackTests.cs#L36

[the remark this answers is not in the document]
andreakarasho: ## Assembly.GetExecutingAssembly()
```csharp
using System;


Console.WriteLine("Assembly.GetExecutingAssembly = {0}", System.Reflection.Assembly.GetExecutingAssembly().Location);
```

### Minimum size
`bflat build --no-reflection --no-stacktrace-data --no-globalization --no-exception-messages --Os`

results:
```
Unhandled Exception: EETypeRva:0x00048E88: Arg_NotImplementedException
   at Test!<BaseAddress>+0xab100
   at Test!<BaseAddress>+0xab121
   at Test!<BaseAddress>+0x9da56
   at Test!<BaseAddress>+0xaa00a
   at Test!<BaseAddress>+0xaa3ac
   at Test!<BaseAddress>+0xb2c1b
   at Test!<BaseAddress>+0xb4520
   at Test!<BaseAddress>+0xbc589
```

### Standard build
`bflat build`

results [seems null/empty]:
```
Assembly.GetExecutingAssembly =
```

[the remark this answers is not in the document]
TheSpydog: Interesting, I just ran a non-bflat NativeAOT build to see what happens. GetExecutingAssembly returns the assembly information, but its Location is null/empty string. I'm kinda wondering if we should just hardcode "FNA.dll.config" for the filepath instead of using the Location string. I think it uses a relative file search path anyway... I can check that later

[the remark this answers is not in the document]
flibitijibibo: A fallback to FNA.dll.config makes sense to me (maybe with a warning, even though it'd print 100% of the time on NativeAOT).

TheSpydog:
I think I've added everything except dllentry at this point. Looking at the spec it appears that only `dllentry` elements can take a `name` attribute, not `dllmap`. Since remapping function names is the sole(?) purpose of dllentry elements and NativeLibrary doesn't support that feature, it doesn't look like dllentry is worth the hassle of implementing to spec. I'm logging a warning if any appear in the config file.

flibitijibibo (approved):

src/Utilities/FNADllMap.cs
```diff
@@ -0,0 +1,184 @@
...
+
+		#region Module Initializer
+
+		[ModuleInitializer]
+		public static void Init()
+		{
+			// Get the platform and architecture
+			string os = GetPlatformName();
+			string cpu = RuntimeInformation.ProcessArchitecture.ToString().ToLowerInvariant();
+			string wordsize = (IntPtr.Size == 4) ? "32" : "64";
```
flibitijibibo, line 87: Nit: (IntPtr.Size * 8).ToString()

src/Utilities/FNADllMap.cs
```diff
@@ -0,0 +1,184 @@
...
+			// Load the XML
+			XmlDocument xmlDoc = new XmlDocument();
+			xmlDoc.Load(xmlPath);
+
+			// The NativeLibrary API cannot remap function names. :(
+			if (xmlDoc.GetElementsByTagName("dllentry") != null)
+			{
+				string msg = "Function remapping is not supported by .NET Core. Ignoring dllentry elements...";
+				Console.WriteLine(msg);
+				System.Diagnostics.Debug.WriteLine(msg); // Log it in the debugger for non-console apps.
```
flibitijibibo, line 107: Nit: Namespace, also maybe do Debugger.IsAttached so it doesn't double print

src/Utilities/FNADllMap.cs
```diff
@@ -0,0 +1,184 @@
...
+
+			// Set the resolver callback
+			NativeLibrary.SetDllImportResolver(assembly, MapAndLoad);
+		}
+
+		#endregion
+	}
+}
+
+#endif
```
flibitijibibo, line 184: Nit: `#endif // NET5_0`

flibitijibibo (commented):
Extremely annoying nitpicks to try and make this as clean and robust as humanly imaginable.

src/Utilities/FNADllMap.cs
```diff
@@ -0,0 +1,184 @@
+#region License
+/* FNA - XNA4 Reimplementation for Desktop Platforms
+ * Copyright 2009-2021 Ethan Lee and the MonoGame Team
+ *
+ * Released under the Microsoft Public License.
+ * See LICENSE for details.
+ */
+#endregion
+
+#if NET5_0
```
andreakarasho, line 10: I would suggest to use the `NET5_0_OR_GREATER` notation as described here: https://github.com/dotnet/designs/blob/main/accepted/2020/or-greater-defines/or-greater-defines.md

TheSpydog: It looks like those defines are only available in .NET 6+, but there is a generic `NET` define that means 5.0+. We'll use that one instead.

lithiumtoast:
Testing using static libraries with `bflat` compiler (nAOT)... There is no need for DllMap in this situation. DllMap is only relevant for shared libraries which use `dlysym`/`GetProcAddress` underneath the hood with `DllImport` attribute.

flibitijibibo:
A quick checklist so anyone invested in this can produce the data needed to verify this for master... build [Flotilla](https://github.com/blendogames/flotilla) or any XNA game that uses both FNA3D and FAudio against the following targets:

- [ ] CoreCLR
    - [ ] Windows (Copy x64 fnalibs to exe directory)
    - [ ] macOS (Add `@executable_path/osx` to rpath!)
    - [ ] Linux (Copy lib64 fnalibs to netcoredeps)
- [ ] NativeAOT
    - [ ] Windows (Copy x64 fnalibs to exe directory)
    - [ ] macOS (Add `@executable_path/osx` to rpath!)
    - [ ] Linux (Copy lib64 fnalibs to netcoredeps? Does NativeAOT set rpath?)

If we can show this is reliable for both JIT and AOT and it respects symversioning correctly I'll merge this in.

andreakarasho:
For the whole NativeAOT section i can share my ImGui game launcher for win, linux, osx.
But no audio used.
You can grab executables from [here](https://www.classicuo.eu/).

In addiction:
i compiled my fna game with bflat on windows days ago.
It runs well with the older DllMap: https://github.com/andreakarasho/ClassicUO/blob/dev/src/DllMap.cs
The relevant structure folder is:
```
x64 [contains all FNA dlls, so i did not copy to the exe origin]
game_compiled_with_bflat.exe
FNA.dll.config
```

flibitijibibo:
It turns out NativeAOT actually has the ability to dynamically link libraries in place of P/Invoke, so we should definitely do that instead as it will produce far more reliable output. With that in mind we should produce the msbuild data needed to make that happen, and instead of supporting DllMap there we can make sure that having this at least doesn't _break_ AOT (if we need to check for a def here that's fine as well).

src/Utilities/FNADllMap.cs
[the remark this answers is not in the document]
MichalStrehovsky: [Assembly.Location will be always null/empty](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/il3000) when deployed as single file because the assembly doesn't exist on the filesystem anymore.

This is also true for CoreCLR single file publishing after .NET 5 (the single file that isn't just a glorified self-extracting ZIP) - it's not specific to NativeAOT single file.

PR #358, opened 2021-07-20
Game loop timing improvements

thatcosmonaut:
1) We were polling events before sleeping, which was causing input latency.

2) Currently our fixed timestep sleep puts us entirely at the mercy of the system scheduler, so now we sleep for 1ms intervals, checking worst case timing the system sleeps for, and making adjustments based on that.

### DIFF
--- a/src/Game.cs
+++ b/src/Game.cs
@@ -212,6 +212,9 @@ namespace Microsoft.Xna.Framework
 		private long previousTicks = 0;
 		private int updateFrameLag;
 		private bool forceElapsedTimeToZero = false;
+		private TimeSpan[] previousSleepTimes = new TimeSpan[100];
+		private int sleepTimeIndex = 0;
+		private TimeSpan worstCaseSleepPrecision = TimeSpan.FromMilliseconds(1);
 
 		private static readonly TimeSpan MaxElapsedTime = TimeSpan.FromMilliseconds(500);
 
@@ -250,6 +253,10 @@ namespace Microsoft.Xna.Framework
 			IsFixedTimeStep = true;
 			TargetElapsedTime = TimeSpan.FromTicks(166667); // 60fps
 			InactiveSleepTime = TimeSpan.FromSeconds(0.02);
+			for (int i = 0; i < previousSleepTimes.Length; i += 1)
+			{
+				previousSleepTimes[i] = TimeSpan.FromMilliseconds(1);
+			}
 
 			textInputControlDown = new bool[FNAPlatform.TextInputCharacters.Length];
 			textInputControlRepeat = new int[FNAPlatform.TextInputCharacters.Length];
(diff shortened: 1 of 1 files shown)


flibitijibibo:
CC @mrhelmut, who is the last person to look at Game.Tick() from what I can tell.

In our local testing this seems to be a dramatic improvement for the hardware we have - timing is very stable and input isn't laggy anymore, and we've given this to some speedrunners to verify. We would like the MonoGame team's input, however, because for the most part we share the core Tick(), so in addition to sharing a really important change with MonoGame it'd be great if both projects could test this with a lot of hardware, since we still don't know how sensitive this is across the various CPUs running games today. We'll continue to test this locally but we would love to be able to share a potentially huge improvement with as many developers and players as possible.

src/Game.cs
```diff
@@ -212,6 +212,9 @@ public GameWindow Window
 		private long previousTicks = 0;
 		private int updateFrameLag;
 		private bool forceElapsedTimeToZero = false;
+		private TimeSpan[] previousSleepTimes = new TimeSpan[100];
```
jaedan, line 215: If you make the array a power of two size, like 64, you can get rid of the expensive mod operation below and switch to a cheap bit mask.

src/Game.cs
```diff
@@ -854,18 +865,38 @@ private void RunLoop()
...
+			previousTicks = currentTicks;
+			return timeAdvanced;
+		}
+
+		/* To calculate the sleep precision of the OS, we take the worst case
+		 * time spent sleeping over the results of previous requests to sleep 1ms.
+		 */
+		private void UpdateEstimatedSleepPrecision(TimeSpan timeSpentSleeping)
+		{
+			previousSleepTimes[sleepTimeIndex] = timeSpentSleeping;
```
jaedan, line 887: You know the previous worst case - it's saved in worstCaseSleepPrecision. You also know your current index. So the only way the worst case changes is if you either 1) just got a new worst case, or 2) the worst case was the oldest entry on the list. You can check that quickly with something like this:

```
if (timeSpentSleeping >= worstCaseSleepPrecision)
{
    worstCaseSleepPrecision = timeSpentSleeping;
}
else if (previousSleepTimes[sleepTimeIndex] == worstCaseSleepTime)
{
    var maxSleepTime = TimeSpan.MinValue;
    for (int i = 0; i < previousSleepTimes.Count; i++)
    {
        if (previousSleepTimes[i] > maxSleepTime)
        {
            maxSleepTime = sleepTime;
        }
    }
    worstCaseSleepPrecision = maxSleepTime;
}

previousSleepTimes[sleepTimeIndex] = timeSpentSleeping;
sleepTimeIndex = (sleepTimeIndex + 1) & SLEEP_TIME_MASK;
```

flibitijibibo: Crap, obvious catch, thanks for finding this >_<

thatcosmonaut: Seconded, thanks for this.

thatcosmonaut:
Moving to https://github.com/FNA-XNA/FNA/pull/359 so we can work from upstream.

PR #359, opened 2021-07-20
Game loop timing improvements

thatcosmonaut:
1) We were polling events before sleeping, which was causing input latency.

2) Currently our fixed timestep sleep puts us entirely at the mercy of the system scheduler, so now we sleep for 1ms intervals, keep track of the worst case timing the system sleeps for, and make adjustments based on that. Then we SpinWait for the final timing interval. So far this approach seems to grant very smooth timing.

Replacement DLLs built with this patch are hosted here: http://moonside.games/files/FNATimingFix.zip

### DIFF
--- a/src/Game.cs
+++ b/src/Game.cs
@@ -212,6 +212,9 @@ namespace Microsoft.Xna.Framework
 		private long previousTicks = 0;
 		private int updateFrameLag;
 		private bool forceElapsedTimeToZero = false;
+		private TimeSpan[] previousSleepTimes = new TimeSpan[100];
+		private int sleepTimeIndex = 0;
+		private TimeSpan worstCaseSleepPrecision = TimeSpan.FromMilliseconds(1);
 
 		private static readonly TimeSpan MaxElapsedTime = TimeSpan.FromMilliseconds(500);
 
@@ -250,6 +253,10 @@ namespace Microsoft.Xna.Framework
 			IsFixedTimeStep = true;
 			TargetElapsedTime = TimeSpan.FromTicks(166667); // 60fps
 			InactiveSleepTime = TimeSpan.FromSeconds(0.02);
+			for (int i = 0; i < previousSleepTimes.Length; i += 1)
+			{
+				previousSleepTimes[i] = TimeSpan.FromMilliseconds(1);
+			}
 
 			textInputControlDown = new bool[FNAPlatform.TextInputCharacters.Length];
 			textInputControlRepeat = new int[FNAPlatform.TextInputCharacters.Length];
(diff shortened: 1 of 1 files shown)


flibitijibibo (approved):
The feedback we've gotten has been positive, but limited. So I'll make the call: This isn't going into 21.08, but will be the very first commit after the tag. Developers will have the entire month to try this out, and we'll bring it up in the 21.08 announcement post.

src/Game.cs
```diff
@@ -212,6 +212,12 @@ public GameWindow Window
 		private long previousTicks = 0;
 		private int updateFrameLag;
 		private bool forceElapsedTimeToZero = false;
+		// must be a power of 2 so we can do a bitmask optimization when checking worst case
```
flibitijibibo, line 215:
```suggestion

		// must be a power of 2 so we can do a bitmask optimization when checking worst case
```

src/Game.cs
```diff
@@ -854,18 +868,62 @@ private void RunLoop()
...
+		/* To calculate the sleep precision of the OS, we take the worst case
+		 * time spent sleeping over the results of previous requests to sleep 1ms.
+		 */
+		private void UpdateEstimatedSleepPrecision(TimeSpan timeSpentSleeping)
+		{
+			/* It is unlikely that the scheduler will actually be more imprecise than
+			 * 4ms and we don't want to get wrecked by a single long sleep so we cap this
+			 * value at 4ms for sanity.
+			 */
+			var upperTimeBound = TimeSpan.FromMilliseconds(4);
```
flibitijibibo, line 894:
```suggestion
			TimeSpan upperTimeBound = TimeSpan.FromMilliseconds(4);
```

src/Game.cs
```diff
@@ -854,18 +868,62 @@ private void RunLoop()
...
+			 * is if we either 1) just got a new worst case, or 2) the worst case was
+			 * the oldest entry on the list.
+			 */
+			if (timeSpentSleeping >= worstCaseSleepPrecision)
+			{
+				worstCaseSleepPrecision = timeSpentSleeping;
+			}
+			else if (previousSleepTimes[sleepTimeIndex] == worstCaseSleepPrecision)
+			{
+				var maxSleepTime = TimeSpan.MinValue;
```
flibitijibibo, line 912:
```suggestion
				TimeSpan maxSleepTime = TimeSpan.MinValue;
```

Also, do we care that we're initializing these inside the function each time? Does `const` make these static inside the function or should we have them as const values at the top with the other new variables?

src/Game.cs
```diff
@@ -854,18 +868,62 @@ private void RunLoop()
...
+				for (int i = 0; i < previousSleepTimes.Length; i++)
```
flibitijibibo, line 913:
```suggestion
				for (int i = 0; i < previousSleepTimes.Length; i += 1)
```

flibitijibibo (changes requested):
Mostly cosmetic stuff coming in at the last minute, though the TimeSpan constructors have me wondering if we can make this a tiny bit faster.

src/Game.cs
thatcosmonaut: TimeSpan is a struct and the constructor doesn't do a whole lot. I think this would be an unnecessary micro-optimization

https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs

PR #362, opened 2021-09-11
Add Srgb render target support

bartwe:


flibitijibibo (changes requested):
Need to undo the submodule update and fix tabbing, otherwise fine

bartwe:
How do you mean 'undo the submodule?' it contains required changes

flibitijibibo:
We have to use the revision that gets merges into the master branch first - I'll merge that first, then we can use that revision for this patch.

src/Graphics/SurfaceFormat.cs
```diff
@@ -98,5 +98,10 @@ public enum SurfaceFormat
 		/// Unsigned 32-bit ABGR pixel format for store 8 bits per channel (XNA3)
 		/// </summary>
 		ColorBgraEXT,
-	}
+		/// <summary>
+		/// Unsigned 32-bit ARGB pixel format for store 8 bits per channel.
+		/// Byte encoding is in sRGB colorspace, read in shader in linear colorspace.
+		/// </summary>
+		ColorSrgbEXT,
+    }
```
flibitijibibo, line 106:
```suggestion
	}
```

src/Graphics/GraphicsAdapter.cs
```diff
@@ -206,6 +206,7 @@ public bool IsProfileSupported(GraphicsProfile graphicsProfile)
 			 * -flibit
 			 */
 			if (	format != SurfaceFormat.Color &&
+				format != SurfaceFormat.ColorSrgbEXT &&
```
flibitijibibo, line 209: Should be at the bottom of the list

flibitijibibo (changes requested):
Couple minor things...

flibitijibibo:
Wrote in the bare minimum since this took too long... should be enough unless someone's hacking into ProfileCapabilities:

https://github.com/FNA-XNA/FNA/commit/7fb834b7ad59d81d2b3eb783f7673fb1d0c6e3e5

PR #371, opened 2021-12-26
Add support for multiple video tracks

kiddkaffeine:


flibitijibibo:
Functionally it makes sense, but let's simplify this by disallowing video streams where size does not match. Theorafile could possibly support it separately, but similar to audio we should assert that the formats should be identical; you should only need to change streams for non-technical reasons (i.e. language changes). For different formats, different files should be used - trying to support multiple formats in one container is too error-prone given the XNA spec's constraints.

kiddkaffeine:
Removed the logic for resizing VideoPlayer mid-stream.  This may be inadvisable anyway; changing video tracks midstream (even when everything is generally almost identical) generally seems to cause momentarily compression artifacting.

src/Media/Xiph/Video.cs
```diff
@@ -173,6 +173,13 @@ public void SetAudioTrackEXT(int track)
 			}
 		}
 
+		public void SetVideoTrackEXT(int track)
+		{
+			if (theora != IntPtr.Zero) {
```
flibitijibibo, line 178:
```suggestion
			if (theora != IntPtr.Zero)
			{
```

flibitijibibo (approved):
Just a newline and this is good

src/Media/Xiph/Video.cs
kiddkaffeine: Updated.

flibitijibibo:
If I don't rebuild tonight I'll likely get this merged on the 30th. If anyone wants to check those sizeof values between now and then, that'll help too.

PR #372, opened 2022-01-09
Add BC7 support

kg:
Depends on https://github.com/FNA-XNA/FNA3D/pull/121

### DIFF
--- a/src/Graphics/SurfaceFormat.cs
+++ b/src/Graphics/SurfaceFormat.cs
@@ -107,6 +107,7 @@ namespace Microsoft.Xna.Framework.Graphics
 		/// DXT5. Texture format with compression. Surface dimensions must be a multiple 4.
 		/// Byte encoding is in sRGB colorspace, read in shader in linear colorspace.
 		/// </summary>
+		Dxt5SrgbEXT,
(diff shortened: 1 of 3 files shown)


src/Graphics/SurfaceFormat.cs
```diff
@@ -108,5 +108,13 @@ public enum SurfaceFormat
 		/// Byte encoding is in sRGB colorspace, read in shader in linear colorspace.
 		/// </summary>
 		Dxt5SrgbEXT,
+		/// <summary>
+		/// BC7 block texture format
+		/// </summary>
+		BC7EXT,
```
flibitijibibo: We have to be goofy and do "Bc7EXT", emphasis on 'c', to stick with the XNA casing scheme. (Also I think Supergiant capitalizes it this way in their C# engine.)

flibitijibibo (approved):
Wouldn't let me delete the comment, this is fine now

kg:
Before we merge do I need to put a FNA3d bump in here? Or will you do it once we merge the FNA3d side?

flibitijibibo:
I'll do it on my end, so nothing to worry about here.

PR #376, opened 2022-01-24
Add @rpath/ prefix to OSX as workaround to changes in ordering

kiddkaffeine:
Tested this change on a Catalina, Big Sur, and Monterey box. Seems to work correctly.


flibitijibibo:
This doesn't cover everything - we also need to apply it to every DllImport, including those without a dllmap. We need to propose this to upstream Mono as the new first search for P/Invoke calls. They have this bug too so they'll be receptive to this even without dllmap, should Apple decide this is a feature and not a bug.

PR #379, opened 2022-01-29
Adding support for float textures to DDS

YTN0:
Needed to load some Cubemaps which were using 64bpp and 128bpp floats and discovered that FNA did not support this. 

I managed to get it working via these changes to Texture.cs.

### DIFF
--- a/src/Graphics/Texture.cs
+++ b/src/Graphics/Texture.cs
@@ -220,7 +220,7 @@ namespace Microsoft.Xna.Framework.Graphics
 			const uint FOURCC_DXT3 = 0x33545844;
 			const uint FOURCC_DXT5 = 0x35545844;
 			const uint FOURCC_BPTC = 0x30315844;
-			//const uint FOURCC_DX10 = 0x30315844;
+			// const uint FOURCC_DX10 = 0x30315844;
 			const uint pitchAndLinear = (
 				DDSD_PITCH | DDSD_LINEARSIZE
 			);
@@ -315,7 +315,7 @@ namespace Microsoft.Xna.Framework.Graphics
 						break;
 					case FOURCC_BPTC:
 						format = SurfaceFormat.Bc7EXT;
-						break;						
+						break;
 					default:
 						throw new NotSupportedException(
 							"Unsupported DDS texture format"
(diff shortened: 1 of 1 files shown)


src/Graphics/Texture.cs
```diff
@@ -212,7 +222,7 @@ SurfaceFormat format
 			const uint FOURCC_DXT3 = 0x33545844;
 			const uint FOURCC_DXT5 = 0x35545844;
 			const uint FOURCC_BPTC = 0x30315844;
-			// const uint FOURCC_DX10 = 0x30315844;
+//			const uint FOURCC_DX10 = 0x30315844;
```
flibitijibibo, line 225: This can be undone...

src/Graphics/Texture.cs
```diff
@@ -166,6 +166,16 @@ SurfaceFormat format
 				return (((width * 32) + 7) / 8) * height;
 			}
 			else
+			if (format == SurfaceFormat.HalfVector4)
```
flibitijibibo, line 169: Interesting pattern but should be `else if` on one line

src/Graphics/Texture.cs
```diff
@@ -290,6 +300,12 @@ SurfaceFormat format
 			{
 				switch (formatFourCC)
 				{
+					case 0x71:
```
flibitijibibo, line 303: Need FOURCC constants for these two

src/Graphics/Texture.cs
```diff
@@ -301,7 +317,7 @@ SurfaceFormat format
 						break;
 					case FOURCC_BPTC:
 						format = SurfaceFormat.Bc7EXT;
-						break;
+						break;
```
flibitijibibo, line 320: Extra whitespace

flibitijibibo (changes requested):
Functionally okay, just needs cleaning up is all

src/Graphics/Texture.cs
YTN0: There are no FOURCC constants for these. These are D3DFORMAT enums. See the following MS code:

https://docs.microsoft.com/en-us/windows/uwp/gaming/complete-code-for-ddstextureloader

flibitijibibo: Got it - let's comment with what the magic numbers mean, then this is good to go.

YTN0: Done!

PR #386, opened 2022-05-12
Fix BC7 loading

thatcosmonaut:


kg:
I think you should verify whether that header is always there. I'm pretty sure the bc7 dds file I got out of renderdoc didn't have it.

thatcosmonaut:
Looks like some encoders do add the extra header and some don't: https://docs.microsoft.com/en-us/windows/win32/direct3ddds/dx-graphics-dds-pguide#dds-file-layout

I'll patch it to support both tomorrow

thatcosmonaut:
Updated the parser based on that Microsoft guide. My test cases work but we should test Kate's input too.

flibitijibibo:
This change looks okay to me, but it's odd that you can either have all that stuff or nothing at all... assuming both types of data are valid I wonder if there's just a length value we've missed?

thatcosmonaut:
The document I linked above lays it out pretty clearly, for DXT5/BC3 pixel format and lower the old DDS header can be used, for anything past that the DX10 extra header flags are required, but newer tools can optionally output the DX10 extra header even for the older pixel formats. If the header has the `DX10` fourCC value then it must have the DX10 extra header, not having that is a spec violation:

> If the DDS_PIXELFORMAT dwFlags is set to DDPF_FOURCC and dwFourCC is set to "DX10" an additional [DDS_HEADER_DXT10](https://docs.microsoft.com/en-us/windows/win32/direct3ddds/dds-header-dxt10) structure will be present to accommodate texture arrays or DXGI formats that cannot be expressed as an RGB pixel format such as floating point formats, sRGB formats etc. When the DDS_HEADER_DXT10 structure is present the entire data description will looks like this.
> 
> DWORD               dwMagic;
> DDS_HEADER          header;
> DDS_HEADER_DXT10    header10;

thatcosmonaut:
Test inputs passed, this is ready for review.

src/Graphics/Texture.cs
```diff
@@ -313,8 +323,80 @@ SurfaceFormat format
...
+						uint resourceDimension = reader.ReadUInt32();
+
+						// These values are taken from the D3D10_RESOURCE_DIMENSION enum.
+						switch (resourceDimension)
+						{
+							case 0: // Unknown
+							case 1: // Buffer
+								throw new NotSupportedException(
+									"Unsupported DDS texture format"
+								);
```
flibitijibibo, line 376: Worth having an empty default handler, genuinely thought this would throw 100% until I stared at it for a minute.

flibitijibibo (commented):
Should be fine minus a visual thing.

src/Graphics/Texture.cs
thatcosmonaut: Done

PR #388, opened 2022-05-30
Fix for repeating non-text input speed issue.

teamtwentythree:
The current code handles non-text repeating inside FNA, however it doesn't delay the repeat which can cause timing issues when running at higher frame rates (Backspace deletes text faster than a user would expect).  Per Ethan's recommendation using the repeat field from the sdl event allows repeating using the expected OS behavior/speed.  

As part of this change I also removed the repeat array which is no longer used.  Additionally added support to paste using right control as well as left, as is expected behavior on Windows (Do not have immediate access to Linux/Mac to test).

This does alter the parameters of public PollEvents function + delegate so could potentially cause issues for folks touching PollEvents (Not sure why anyone would).

### DIFF
--- a/src/FNAPlatform/SDL2_FNAPlatform.cs
+++ b/src/FNAPlatform/SDL2_FNAPlatform.cs
@@ -875,7 +875,7 @@ namespace Microsoft.Xna.Framework
 						}
 					}
 					else if (evt.key.repeat > 0)
-                    {
+					{
 						int textIndex;
 						if (FNAPlatform.TextInputBindings.TryGetValue(key, out textIndex))
 						{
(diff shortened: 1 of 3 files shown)


src/FNAPlatform/SDL2_FNAPlatform.cs
```diff
@@ -865,17 +864,29 @@ public static void UnregisterGame(Game game)
...
+							&& key == Keys.V)
 						{
 							textInputControlDown[6] = true;
-							textInputControlRepeat[6] = Environment.TickCount + 400;
 							TextInputEXT.OnTextInput(FNAPlatform.TextInputCharacters[6]);
 							textInputSuppress = true;
 						}
 					}
+					else if (evt.key.repeat > 0)
+                    {
```
flibitijibibo, line 878: Text editor trolled us:

```suggestion
					{
```

flibitijibibo (commented):
Just the one line, everything else is fine!

PR #394, opened 2022-09-07
Improve SpriteBatch efficiency

CXUtk:
I have changed the resize policy to doubling, which is aligned with XNA implementation. This is to avoid the O(n^2) reallocation process when we are pushing a large number of sprites. Now it only takes O(n) in any situation. In addition to that, I set the limit to a fixed number to avoid using too much memory.

### DIFF
--- a/src/Graphics/SpriteBatch.cs
+++ b/src/Graphics/SpriteBatch.cs
@@ -1090,12 +1090,6 @@ namespace Microsoft.Xna.Framework.Graphics
 					 * increase the graphics buffer sizes!
 					 * -flibit
 					 */
-
-					 /* Fix: reallocation size now increase exponentially, 
-					  * but upper bounded to 1048576. When we submit more than 
-					  * 1048576 sprites, we flush them immediately. This will make
-					  * the total running time O(n).
-					 */ 
 					int newMax = Math.Min(vertexInfo.Length * 2, MAX_ARRAYSIZE);
 					Array.Resize(ref vertexInfo, newMax);
 					Array.Resize(ref textureInfo, newMax);
--- a/src/Graphics/SpriteBatch.cs
+++ b/src/Graphics/SpriteBatch.cs
@@ -30,6 +30,7 @@ namespace Microsoft.Xna.Framework.Graphics
 		private const int MAX_SPRITES = 2048;
 		private const int MAX_VERTICES = MAX_SPRITES * 4;
 		private const int MAX_INDICES = MAX_SPRITES * 6;
+		private const int MAX_BATCHSIZE = 1048576;
 
 		// Used to quickly flip text for DrawString
 		private static readonly float[] axisDirectionX = new float[]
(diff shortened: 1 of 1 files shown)


CXUtk:
> 
1048576 is just 2^20, I set this number somewhat arbitrarily, but you can set it to any upper bound that makes sense for the array. For example, some value that balances the number of draw calls emitted and memory consumption.

flibitijibibo:
Let's do this: The max buffer size is 0x3FFFFFF, so let's use that combined with the size of the vertex stride to pick the largest buffer size that SpriteBatch can make without going over the HiDef limit. Then we can document it as such and this should be good to go.

CXUtk:
It seems like now we only send to GPU 2048 sprites per draw call, but the limit is only related to CPU side, do you want me also change the VertexBuffer size?

src/Graphics/SpriteBatch.cs
```diff
@@ -31,6 +31,9 @@ public class SpriteBatch : GraphicsResource
 		private const int MAX_VERTICES = MAX_SPRITES * 4;
 		private const int MAX_INDICES = MAX_SPRITES * 6;
 
+		// Max array size for cpu side buffer
+		private const int MAX_ARRAYSIZE = 0x3FFFFFF;
```
flibitijibibo, line 35:
```suggestion
		private const int MAX_ARRAYSIZE = 0x3FFFFFF / 96;
```

flibitijibibo (changes requested):
This change combined with comment updates should make this ready to go.

src/Graphics/SpriteBatch.cs
```diff
@@ -1067,17 +1070,30 @@ byte effects
...
+					 * sprite sorting accuracy; note that we do NOT
+					 * increase the graphics buffer sizes!
+					 * -flibit
+					 */
+
+					 /* Fix: reallocation size now increase exponentially, 
+					  * but upper bounded to 1048576. When we submit more than 
+					  * 1048576 sprites, we flush them immediately. This will make
+					  * the total running time O(n).
+					 */
```
flibitijibibo, line 1090:
```suggestion
```

src/Graphics/SpriteBatch.cs
```diff
@@ -31,6 +31,9 @@ public class SpriteBatch : GraphicsResource
 		private const int MAX_VERTICES = MAX_SPRITES * 4;
 		private const int MAX_INDICES = MAX_SPRITES * 6;
 
+		// Max array size for cpu side buffer
```
flibitijibibo, line 34:
```suggestion
		/* This is the largest array size for a VertexBuffer using VertexPositionColorTexture.
		 * note that we do NOT change the GPU buffer size since that would break XNA accuracy,
		 * but if you want to optimize your batching you can make this change in a custom SpriteBatch.
		 */
```

src/Graphics/SpriteBatch.cs
```diff
@@ -1067,17 +1070,30 @@ byte effects
...
-				 * -flibit
-				 */
-				int newMax = vertexInfo.Length + MAX_SPRITES;
-				Array.Resize(ref vertexInfo, newMax);
-				Array.Resize(ref textureInfo, newMax);
-				Array.Resize(ref spriteInfos, newMax);
-				Array.Resize(ref sortedSpriteInfos, newMax);
+				if (vertexInfo.Length >= MAX_ARRAYSIZE)
+				{
+					FlushBatch();
```
flibitijibibo, line 1075:
```suggestion
					/* FIXME: We're doing this for safety but it's possible that
					 * XNA just keeps expanding and crashes with OutOfMemory.
					 * Since GraphicsProfile has a buffer cap, we use that for safety.
					 * This might change if someone depends on running out of memory(?!).
					 */
					FlushBatch();
```

flibitijibibo (changes requested):

flibitijibibo:
Forgot to mention: Was able to verify the resize logic by hitting the limit exactly twice; I drew enough to make it resize and then resize again, and it doubled instead of increasing by the original batch size, so that part is definitely accurate (good catch).

src/Graphics/SpriteBatch.cs
```diff
@@ -31,6 +31,12 @@ public class SpriteBatch : GraphicsResource
 		private const int MAX_VERTICES = MAX_SPRITES * 4;
 		private const int MAX_INDICES = MAX_SPRITES * 6;
 
+		/* This is the largest array size for a VertexBuffer using VertexPositionColorTexture.
+		 * note that we do NOT change the GPU buffer size since that would break XNA accuracy,
```
flibitijibibo, line 35:
```suggestion
		 * Note that we do NOT change the GPU buffer size since that would break XNA accuracy,
```

PR #396, opened 2022-09-20
Add ToString to Touch classes that lack it

kiddkaffeine:
This mainly came up when testing with GestureSample, but I went ahead and added it for any non-static/non-enum class in this namespace lacking it.


flibitijibibo:
So, bad news: Turns out XNA doesn't implement these, even though they absolutely should... sadly for us this means we can't fix this.

Code itself lgtm though, just has to be done in the application instead :/

PR #397, opened 2022-09-27
Add SaveAsJpegEXT that allows quality selection

kiddkaffeine:


kiddkaffeine:
I still believe we should find a way to accommodate this, but it sounds like this is not the preferred approach for now.

PR #402, opened 2022-10-23
Add TextInputEXT.IsScreenKeyboardShown

tomcashman:
SDL's IsTextInputActive() can still return true if on-screen keyboard is closed as this function just checks the SDL event state.

On some platforms (e.g. Xbox), the on-screen keyboard can be closed by an external event which results in the event state being enabled despite the keyboard being no longer visible.

To workaround this, I've added a binding to SDL's IsScreenKeyboardShown to allow for checking this.

Note: On Xbox UWP, there is a bug in SDL which I've submitted PR for [here](https://github.com/libsdl-org/SDL/issues/6433).

### DIFF
--- a/src/Input/TextInputEXT.cs
+++ b/src/Input/TextInputEXT.cs
@@ -37,9 +37,22 @@ namespace Microsoft.Xna.Framework.Input
 
 		#region Public Static Methods
 
+		/// <summary>
+		/// Returns if text input state is active
+		///
+		/// Note: For on-screen keyboard, this may remain true on
+		/// some platforms if an external event closed the keyboard.
+		/// In this case, check IsScreenKeyboardShow instead.
+		/// </summary>
+		/// <returns>True if text input state is active</returns>
 		public static bool IsTextInputActive()
 		{
 			return FNAPlatform.IsTextInputActive();
+		}
+
+		public static bool IsScreenKeyboardShown(IntPtr window)
+		{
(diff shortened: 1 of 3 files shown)


src/Input/TextInputEXT.cs
```diff
@@ -37,11 +37,28 @@ public static class TextInputEXT
...
+		/// some platforms if an external event closed the keyboard.
+		/// In this case, check IsScreenKeyboardShow instead.
+		/// </summary>
+		/// <returns>True if text input state is active</returns>
 		public static bool IsTextInputActive()
 		{
 			return FNAPlatform.IsTextInputActive();
 		}
 
+		public static bool IsScreenKeyboardShown(GameWindow window)
```
flibitijibibo, line 53: Let's make this the direct `IntPtr`, so that the API doesn't depend directly on the Game classes.

flibitijibibo (commented):
Another minor change, otherwise good

src/Input/TextInputEXT.cs
tomcashman: I was originally thinking that but GameWindow doesn't expose the IntPtr to the window. Not sure how this would be passed in from the developer's side?

flibitijibibo: Game.Window.Handle will have it!

tomcashman: Ah yes. Ok will update PR tomorrow.

PR #403, opened 2022-12-13
Fixes for NativeAOT warnings

jaedan:
Minor fixes discovered by running the AOT analyzers. This doesn't fix every warning because not everything is fixable. For example, the content pipeline is heavily based on reflection. I could have annotated all of those APIs to remove the warnings from FNA and shift them into game code calling those APIs, but it's a considerable number of annotations introduced throughout the code base and I wasn't ready to propose cluttering everything up with those just yet.

### DIFF
--- a/FNA.csproj
+++ b/FNA.csproj
@@ -402,6 +402,7 @@
     <Compile Include="src\Utilities\FileHelpers.cs" />
     <Compile Include="src\Utilities\FNADllMap.cs" />
     <Compile Include="src\Utilities\FNAInternalExtensions.cs" />
+    <Compile Include="src\Utilities\MarshalHelper.cs" />
     <Compile Include="src\Utilities\XamarinHelper.cs" />
     <Compile Include="src\Vector2.cs" />
     <Compile Include="src\Vector3.cs" />
--- a/Makefile
+++ b/Makefile
@@ -315,6 +315,7 @@ SRC = \
 	src/Utilities/AssemblyHelper.cs \
 	src/Utilities/FileHelpers.cs \
 	src/Utilities/FNAInternalExtensions.cs \
+	src/Utilities/MarshalHelper.cs \
 	src/Utilities/XamarinHelper.cs \
 	src/Vector2.cs \
 	src/Vector3.cs \
--- a/src/Utilities/FNADllMap.cs
+++ b/src/Utilities/FNADllMap.cs
@@ -96,15 +96,10 @@ namespace Microsoft.Xna.Framework
 
 			// Get the path to the assembly
 			Assembly assembly = Assembly.GetExecutingAssembly();
-			string assemblyPath = "";
-			if (!String.IsNullOrEmpty(assembly.Location))
-			{
-				assemblyPath = Path.GetDirectoryName(assembly.Location);
-			}
 
 			// Locate the config file
 			string xmlPath = Path.Combine(
-				assemblyPath,
+				AppContext.BaseDirectory,
 				assembly.GetName().Name + ".dll.config"
 			);
 			if (!File.Exists(xmlPath))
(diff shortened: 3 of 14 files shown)


flibitijibibo:
For SizeOf let's do what we did in SDL2# and make a function to separate the two versions. This will probably go in the Utilities folder somehow.

Someone else will need to review the Core dllmap changes.

jaedan:
Which version of .net is missing support for Marshal.SizeOf<T>? I tried on .net framework 4.8, .net 5, and .net 7 and they all worked, so I need to find the right set up to reproduce the problem to ensure I've fixed it.

flibitijibibo:
4.0 is the target you want - it has SizeOf but not SizeOf<T>.

thatcosmonaut:
If you're using the newer Core API anyway, Unsafe.SizeOf<T> is preferable to Marshal. `sizeof` operator can also be used for pure value type structs.

jaedan:
But don't these uses really want the unmanaged size? Marshal.SizeOf and Unsafe.SizeOf aren't strictly the same thing and since these are for interop, Marshal.SizeOf I think is the correct choice, even if for all of these types they happen to return the same size.

thatcosmonaut:
You're right, I forgot that certain C# primitives have different sizes from the C API. Never mind!

FNA.csproj
```diff
@@ -402,6 +402,7 @@
     <Compile Include="src\Utilities\FileHelpers.cs" />
     <Compile Include="src\Utilities\FNADllMap.cs" />
     <Compile Include="src\Utilities\FNAInternalExtensions.cs" />
+    <Compile Include="src\Utitlies\MarshalHelper.cs" />
```
flibitijibibo, line 405:
```suggestion
    <Compile Include="src\Utilities\MarshalHelper.cs" />
```
This should probably be done in Makefile as well.

flibitijibibo (commented):
Minor fix for the Framework variant. Other than that this looks okay to me, someone will need to review to DllMap change before this gets approved.

src/Utilities/FNADllMap.cs
```diff
@@ -96,15 +96,10 @@ public static void Init()
...
-			string assemblyPath = "";
-			if (!String.IsNullOrEmpty(assembly.Location))
-			{
-				assemblyPath = Path.GetDirectoryName(assembly.Location);
-			}
 
 			// Locate the config file
 			string xmlPath = Path.Combine(
-				assemblyPath,
+				AppContext.BaseDirectory,
```
kant2002, line 102: That would work only starting from Net 4.6 https://learn.microsoft.com/en-us/dotnet/api/system.appcontext.basedirectory?view=net-7.0

probably should be gated with ifdef

jaedan: This whole file is in an #if NET, so it's .net 5 and newer only here.

FNA.csproj
jaedan: Fixed

jaedan:
Addressed feedback. Should be good to go now.

Makefile
```diff
@@ -315,6 +315,7 @@ SRC = \
 	src/Utilities/AssemblyHelper.cs \
 	src/Utilities/FileHelpers.cs \
 	src/Utilities/FNAInternalExtensions.cs \
+	src/Utilities/MarshalHelper.cs
```
flibitijibibo, line 318:
```suggestion
	src/Utilities/MarshalHelper.cs \
```

flibitijibibo (commented):
One-character fix, otherwise ready to go.

Makefile
jaedan: Can you tell I'm not set up to run the makefile? I'm going to to try run this locally

src/Utilities/FNADllMap.cs
kant2002: Then whole change is ok. `AppContext.BaseDirectory` is preferred way to get location of application files and even works with single-file apps.

jaedan:
Ok, tested on Ubuntu - makefile produces FNA.dll successfully.

PR #417, opened 2023-04-03
Add FromUriEXT method to Video

kiddkaffeine:


src/Media/Xiph/Video.cs
```diff
@@ -164,6 +165,27 @@ VideoSoundtrackType soundtrackType
...
+						"Only local file URIs are supported for now"
+					);
+				}
+				path = uri.LocalPath;
+			} else {
+				path = Path.Combine(
+					TitleLocation.Path,
+					uri.ToString()
+				);
+			}
```
flibitijibibo, line 185:
```suggestion
			if (uri.IsAbsoluteUri)
			{
				// If it's absolute, be sure we can actually get it...
				if (!uri.IsFile)
				{
					throw new InvalidOperationException(
						"Only local file URIs are supported for now"
					);
				}
				path = uri.LocalPath;
			}
			else
			{
				path = Path.Combine(
					TitleLocation.Path,
					uri.ToString()
				);
			}
```

flibitijibibo (approved):
Just me being fussy about formatting, otherwise lgtm

PR #427, opened 2023-08-13
Added FNA_IOS_HOME_INDICATOR for controlling the home indicator

razziefox:
Added an environment variable to control how the home indicator in iOS behaves using SDL_SetHint.

IMG_1067
IMG_1068


flibitijibibo:
On the surface this works, but SDL goes a bit deeper and listens to this hint so that the setting can change at runtime:

https://github.com/libsdl-org/SDL/blob/main/src/video/uikit/SDL_uikitviewcontroller.m#L114

We don't have a system like this, so the current recommendation is to use SDL_SetHint directly.

PR #434, opened 2023-09-24
Reduce EffectParameter-related heap allocations

kg:
This PR optimizes the case where an EffectAnnotationCollection is empty by using a shared 'Empty' instance instead of allocating a bunch of unique ones.

Most Effect techniques only have a single pass, so this PR adds a single-item mode to EffectPassCollection, allowing us to skip allocating the backing List unless someone calls .GetEnumerator on the collection.

In various circumstances we allocate EffectParameter instances and associated collections that aren't actually used, so this PR attempts to allocate them lazily on-demand instead.

These optimizations combined reduce the number of heap objects in my game at a fixed point from 121.2K to 100.5K, and the heap size from 66.2MB to 64.6MB. This should also help reduce GC pause times.

### DIFF
--- a/src/Graphics/Effect/Effect.cs
+++ b/src/Graphics/Effect/Effect.cs
@@ -1018,6 +1018,9 @@ namespace Microsoft.Xna.Framework.Graphics
 			IntPtr rawAnnotations,
 			uint numAnnotations
 		) {
+			if (numAnnotations == 0)
+				return EffectAnnotationCollection.Empty;
+
 			MOJOSHADER_effectAnnotation* annoPtr = (MOJOSHADER_effectAnnotation*) rawAnnotations;
 			List<EffectAnnotation> annotations = new List<EffectAnnotation>((int) numAnnotations);
 			for (int i = 0; i < numAnnotations; i += 1)
--- a/src/Graphics/Effect/Effect.cs
+++ b/src/Graphics/Effect/Effect.cs
@@ -990,7 +990,7 @@ namespace Microsoft.Xna.Framework.Graphics
 				return null;
 			}
 
-			var type = *(MOJOSHADER_symbolTypeInfo*)_type;
+			var type = *(MOJOSHADER_symbolTypeInfo*) _type;
 			EffectParameterCollection structMembers = null;
 			if (type.member_count > 0)
 			{
--- a/src/Graphics/Effect/Effect.cs
+++ b/src/Graphics/Effect/Effect.cs
@@ -986,7 +986,9 @@ namespace Microsoft.Xna.Framework.Graphics
 			IntPtr _type
 		) {
 			if (_type == IntPtr.Zero)
+			{
 				return null;
+			}
 
 			var type = *(MOJOSHADER_symbolTypeInfo*)_type;
 			EffectParameterCollection structMembers = null;
--- a/src/Graphics/Effect/Effect.cs
+++ b/src/Graphics/Effect/Effect.cs
@@ -925,40 +925,6 @@ namespace Microsoft.Xna.Framework.Graphics
 					continue;
 				}
 
-				EffectParameterCollection structMembers = null;
-				if (param.value.type.member_count > 0)
-				{
-					List<EffectParameter> memList = new List<EffectParameter>();
-					unsafe
-					{
-						MOJOSHADER_symbolStructMember* mem = (MOJOSHADER_symbolStructMember*) param.value.type.members;
-						IntPtr curOffset = IntPtr.Zero;
-						for (int j = 0; j < param.value.type.member_count; j += 1)
-						{
-							uint memSize = mem[j].info.rows * mem[j].info.columns;
-							if (mem[j].info.elements > 0)
-							{
-								memSize *= mem[j].info.elements;
-							}
-							memList.Add(new EffectParameter(
-								Marshal.PtrToStringAnsi(mem[j].name),
-								null,
-								(int) mem[j].info.rows,
-								(int) mem[j].info.columns,
-								(int) mem[j].info.elements,
-								XNAClass[(int) mem[j].info.parameter_class],
-								XNAType[(int) mem[j].info.parameter_type],
-								null, // FIXME: Nested structs! -flibit
-								null,
-								param.value.values + curOffset.ToInt32(),
-								memSize * 4
-							));
-							curOffset += (int) memSize * 4;
-						}
-					}
-					structMembers = new EffectParameterCollection(memList);
-				}
-
 				parameters.Add(new EffectParameter(
 					Marshal.PtrToStringAnsi(param.value.name),
 					Marshal.PtrToStringAnsi(param.value.semantic),
@@ -967,7 +933,7 @@ namespace Microsoft.Xna.Framework.Graphics
 					(int) param.value.type.elements,
 					XNAClass[(int) param.value.type.parameter_class],
 					XNAType[(int) param.value.type.parameter_type],
-					structMembers,
+					new IntPtr(&paramPtr[i].value.type),
 					INTERNAL_readAnnotations(
 						param.annotations,
 						param.annotation_count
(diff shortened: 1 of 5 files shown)


src/Graphics/Effect/EffectParameter.cs
```diff
@@ -43,6 +43,12 @@ public int ColumnCount
 			private set;
 		}
 
+		public int ElementCount
```
flibitijibibo, line 46: We'll have to make this internal, since I don't think it's in the public API.

flibitijibibo (commented):
Other than whitespacing/formatting this looks fine, ElementCount we probably can't expose directly but it could be something to make easy to grab via reflection or something.

src/Graphics/Effect/EffectParameter.cs
kg: Yeah I'm fine with making it internal. We could also make it EXT, right?

flibitijibibo: Let's make it internal, since that way we can make it a plain variable without the getter/setter in the way.

kg:
Do you want me to fix the whitespacing/formatting? I'd have to do it by hand since VS seems unwilling to

flibitijibibo:
I'll run through in just a moment...

kg:
I managed to get the editorconfig close to FNA's style. It mangles Effect.cs though so that one will need to be hand cleaned up, but I'll try to get the rest

src/Graphics/Effect/Effect.cs
```diff
@@ -1014,10 +981,74 @@ private unsafe void INTERNAL_parseEffectStruct(IntPtr effectData)
 			Techniques = new EffectTechniqueCollection(techniques);
 		}
 
+		internal unsafe static EffectParameterCollection INTERNAL_readEffectParameterStructureMembers(
+			EffectParameter parameter,
+			IntPtr _type
+		) {
+			if (_type == IntPtr.Zero)
+				return null;
```
flibitijibibo, line 989:
```suggestion
			if (_type == IntPtr.Zero)
			{
				return null;
			}
```

src/Graphics/Effect/Effect.cs
```diff
@@ -1014,10 +981,74 @@ private unsafe void INTERNAL_parseEffectStruct(IntPtr effectData)
...
+
+			var type = *(MOJOSHADER_symbolTypeInfo*)_type;
```
flibitijibibo, line 991:
```suggestion
			var type = *(MOJOSHADER_symbolTypeInfo*) _type;
```

src/Graphics/Effect/Effect.cs
```diff
@@ -1014,10 +981,74 @@ private unsafe void INTERNAL_parseEffectStruct(IntPtr effectData)
...
+				index
+			);
+		}
+
 		private unsafe EffectAnnotationCollection INTERNAL_readAnnotations(
 			IntPtr rawAnnotations,
 			uint numAnnotations
 		) {
+			if (numAnnotations == 0)
+				return EffectAnnotationCollection.Empty;
```
flibitijibibo, line 1050:
```suggestion
			if (numAnnotations == 0)
			{
				return EffectAnnotationCollection.Empty;
			}
```

src/Graphics/Effect/EffectParameter.cs
```diff
@@ -57,14 +57,22 @@ public EffectParameterType ParameterType
 
 		public EffectParameterCollection Elements
 		{
-			get;
-			private set;
+			get
+			{
+				if ((elementCount > 0) && (elements == null))
+					BuildElementList();
```
flibitijibibo, line 63:
```suggestion
				if ((elementCount > 0) && (elements == null))
				{
					BuildElementList();
				}
```

src/Graphics/Effect/EffectParameter.cs
```diff
@@ -57,14 +57,22 @@ public EffectParameterType ParameterType
...
 		}
 
 		public EffectParameterCollection StructureMembers
 		{
-			get;
-			private set;
+			get
+			{
+				if ((mojoType != IntPtr.Zero) && (members == null))
+					BuildMemberList();
```
flibitijibibo, line 73:
```suggestion
				if ((mojoType != IntPtr.Zero) && (members == null))
				{
					BuildMemberList();
				}
```

src/Graphics/Effect/EffectParameter.cs
```diff
@@ -108,10 +153,30 @@ uint dataSizeBytes
...
+			ParameterType = parameterType;
+			members = structureMembers;
+			Annotations = annotations;
+			values = data;
+			valuesSizeBytes = dataSizeBytes;
+		}
+
+		#endregion
+
+		#region Allocation optimizations
```
flibitijibibo, line 167:
```suggestion
		#region Allocation Optimizations

```

src/Graphics/Effect/EffectPassCollection.cs
```diff
@@ -22,22 +23,38 @@ public int Count
 		{
 			get
 			{
-				return elements.Count;
+				if (elements == null)
+					return singleItem != null ? 1 : 0;
+				else
+					return elements.Count;
```
flibitijibibo, line 29:
```suggestion
				if (elements == null)
				{
					return singleItem != null ? 1 : 0;
				}
				return elements.Count;
```

src/Graphics/Effect/EffectPassCollection.cs
```diff
@@ -22,22 +23,38 @@ public int Count
...
 		public EffectPass this[int index]
 		{
 			get
 			{
-				return elements[index];
+				if (elements != null)
+					return elements[index];
+
+				if (index != 0)
+					throw new ArgumentOutOfRangeException("index");
```
flibitijibibo, line 41:
```suggestion
				if (elements != null)
				{
					return elements[index];
				}

				if (index != 0)
				{
					throw new ArgumentOutOfRangeException("index");
				}
```

src/Graphics/Effect/EffectPassCollection.cs
```diff
@@ -22,22 +23,38 @@ public int Count
...
 		public EffectPass this[string name]
 		{
 			get
 			{
+				if (elements == null)
+				{
+					if (singleItem.Name.Equals(name))
+						return singleItem;
+					else
+						return null;
```
flibitijibibo, line 55:
```suggestion
					if (singleItem.Name.Equals(name))
					{
						return singleItem;
					}
					return null;
```

src/Graphics/Effect/EffectPassCollection.cs
```diff
@@ -64,13 +82,30 @@ internal EffectPassCollection(List<EffectPass> value)
...
 		}
 
+		internal EffectPassCollection(EffectPass pass)
+		{
+			singleItem = pass;
+		}
+
+		#endregion
+
+		#region Allocation optimization
```
flibitijibibo, line 92:
```suggestion
		#region Allocation Optimization

```

src/Graphics/Effect/EffectPassCollection.cs
```diff
@@ -64,13 +82,30 @@ internal EffectPassCollection(List<EffectPass> value)
...
+		internal List<EffectPass> GetList()
+		{
+			if (elements == null)
+			{
+				elements = new List<EffectPass>(1);
+				elements.Add(singleItem);
+			}
+			return elements;
+		}
```
flibitijibibo, line 101:
```suggestion
		}

```

flibitijibibo (approved):
Just a bunch of little guys, once this is in it can be merged. If you want to rebase the original 3 patches let me know and I'll hold off, otherwise we can just squash it all down.

flibitijibibo (approved):
Latest lgtm, will wait on merging to give time for rebasing and/or additional review, since this does touch our shader system it wouldn't hurt to have more eyes on this...

Can definitely merge the editorconfig changes right away as a separate PR if that simplifies things.

src/Graphics/Effect/EffectAnnotationCollection.cs
```diff
@@ -16,6 +16,11 @@ namespace Microsoft.Xna.Framework.Graphics
 {
 	public sealed class EffectAnnotationCollection : IEnumerable<EffectAnnotation>, IEnumerable
 	{
+		#region Allocation optimization
+		internal static readonly EffectAnnotationCollection Empty =
+			new EffectAnnotationCollection(new List<EffectAnnotation>());
+		#endregion
```
flibitijibibo, line 22:
```suggestion
		#region Allocation optimization

		internal static readonly EffectAnnotationCollection Empty =
			new EffectAnnotationCollection(new List<EffectAnnotation>());

		#endregion
```

src/Graphics/Effect/EffectParameter.cs
```diff
@@ -151,30 +221,23 @@ uint dataSizeBytes
...
 			}
-			ParameterClass = parameterClass;
-			ParameterType = parameterType;
-			StructureMembers = structureMembers;
-			Annotations = annotations;
-			values = data;
-			valuesSizeBytes = dataSizeBytes;
 		}
-
 		#endregion
```
flibitijibibo, line 241:
```suggestion

		#endregion
```

flibitijibibo (approved):
Merging in a sec, just pushing a couple last-minute bits

PR #437, opened 2023-09-28
Replace FNA3D and SDL SetEnvironmentVariable calls with SDL_SetHintWithPriority

### LINKED ISSUE #435: SetEnvironmentVariable has no effect on Linux with .NET Core
Apparently a well known compatibility issue with .NET and `getenv`. SDL can only see the environment of the process when it was loaded, and not any managed changes.

This makes all the calls to `Environment.SetEnvironmentVariable` which set variables for `FNA3D_` or `SDL_` ineffective on Linux with .NET Core.

Remedy is to use `SDL_SetHintWithPriority(..., SDL_HINT_OVERRIDE)`

Note that this issue does *not* affect Mono.

Test code:
```cs
SDL2.SDL.SDL_GetVersion(out var ver);
Console.WriteLine($"SDL Version {ver.major}.{ver.minor}.{ver.patch}");

Console.WriteLine("GetHint(SDL_VIDEODRIVER): " + SDL2.SDL.SDL_GetHint("SDL_VIDEODRIVER"));

Console.WriteLine("SetEnvironmentVariable(SDL_VIDEODRIVER, proc)");
Environment.SetEnvironmentVariable("SDL_VIDEODRIVER", "proc");

Console.WriteLine("GetHint(SDL_VIDEODRIVER): " + SDL2.SDL.SDL_GetHint("SDL_VIDEODRIVER"));

Console.WriteLine("SDL_SetHintWithPriority(SDL_VIDEODRIVER, ovr, SDL_HINT_OVERRIDE)");
SDL2.SDL.SDL_SetHintWithPriority("SDL_VIDEODRIVER", "ovr", SDL2.SDL.SDL_HintPriority.SDL_HINT_OVERRIDE);

Console.WriteLine("GetHint(SDL_VIDEODRIVER): " + SDL2.SDL.SDL_GetHint("SDL_VIDEODRIVER"));
```

Expected output, running with `SDL_VIDEODRIVER=env ...`
```
SDL Version 2.28.2
GetHint(SDL_VIDEODRIVER): env
SetEnvironmentVariable(SDL_VIDEODRIVER, proc)
GetHint(SDL_VIDEODRIVER): proc
SDL_SetHintWithPriority(SDL_VIDEODRIVER, ovr, SDL_HINT_OVERRIDE)
GetHint(SDL_VIDEODRIVER): ovr
```

.NET Core output
```
SDL Version 2.28.2
GetHint(SDL_VIDEODRIVER): env
SetEnvironmentVariable(SDL_VIDEODRIVER, proc)
GetHint(SDL_VIDEODRIVER): env
SDL_SetHintWithPriority(SDL_VIDEODRIVER, ovr, SDL_HINT_OVERRIDE)
GetHint(SDL_VIDEODRIVER): ovr
```

Chicken-Bones:
Fixes #435

### Alternatives
- `FNAPlatform` appears intended to be backend agnostic. The fact that `FNA3D` uses `SDL_GetHint` to read env vars is an implementation detail, so `FNA3D` should probably expose a `SetHint` or `SetEnv` method. (selected)
- ~~A wrapper method could be created which actually calls `setenv` on linux systems.~~

### Backwards Compatibility

Technically a breaking change for anyone using `GetEnvironmentVariable` to read the value of a variable set by FNA for FNA3D/SDL. Note that `FNA_` variables still use `Get/SetEnvironmentVariable` for this reason, as they run in purely managed code.

### DIFF
--- a/src/FNAPlatform/FNAPlatform.cs
+++ b/src/FNAPlatform/FNAPlatform.cs
@@ -38,6 +38,8 @@ namespace Microsoft.Xna.Framework
 
 			// Environment.GetEnvironmentVariable("FNA_PLATFORM_BACKEND");
 
+			SetEnv = SDL2_FNAPlatform.SetEnv;
+
 			// Built-in command line arguments
 			LaunchParameters args = new LaunchParameters();
 			string arg;
@@ -93,7 +95,6 @@ namespace Microsoft.Xna.Framework
 
 			Malloc =			SDL2_FNAPlatform.Malloc;
 			Free =				SDL2.SDL.SDL_free;
-			SetEnv =			SDL2_FNAPlatform.SetEnv;
 			CreateWindow =			SDL2_FNAPlatform.CreateWindow;
 			DisposeWindow =			SDL2_FNAPlatform.DisposeWindow;
 			ApplyWindowChanges =		SDL2_FNAPlatform.ApplyWindowChanges;
--- a/src/FNAPlatform/FNAPlatform.cs
+++ b/src/FNAPlatform/FNAPlatform.cs
@@ -50,30 +50,34 @@ namespace Microsoft.Xna.Framework
 			}
 			if (args.TryGetValue("gldevice", out arg))
 			{
-				Environment.SetEnvironmentVariable(
+				SDL2.SDL.SDL_SetHintWithPriority(
 					"FNA3D_FORCE_DRIVER",
-					arg
+					arg,
+					SDL2.SDL.SDL_HintPriority.SDL_HINT_OVERRIDE
 				);
 			}
 			if (args.TryGetValue("enablelateswaptear", out arg) && arg == "1")
 			{
-				Environment.SetEnvironmentVariable(
+				SDL2.SDL.SDL_SetHintWithPriority(
 					"FNA3D_ENABLE_LATESWAPTEAR",
-					"1"
+					"1",
+					SDL2.SDL.SDL_HintPriority.SDL_HINT_OVERRIDE
 				);
 			}
 			if (args.TryGetValue("mojoshaderprofile", out arg))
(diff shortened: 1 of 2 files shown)


flibitijibibo:
Let's go with option 1, to avoid making SDL calls outside of the SDL2 file.

src/FNAPlatform/FNAPlatform.cs
```diff
@@ -93,6 +93,7 @@ static FNAPlatform()
 
 			Malloc =			SDL2_FNAPlatform.Malloc;
 			Free =				SDL2.SDL.SDL_free;
+			SetEnv =			SDL2_FNAPlatform.SetEnv;
```
flibitijibibo, line 96: Looks like we'll have to set this above the args checks, otherwise we'll get null references.

src/FNAPlatform/SDL2_FNAPlatform.cs
```diff
@@ -385,6 +395,11 @@ public static IntPtr Malloc(int size)
 
 		#endregion
 
+		public static void SetEnv(string name, string value)
+		{
+			SDL.SDL_SetHintWithPriority(name, value, SDL.SDL_HintPriority.SDL_HINT_OVERRIDE);
+		}
+
```
flibitijibibo, line 402:
```suggestion
		#region Environment Variables

		public static void SetEnv(string name, string value)
		{
			SDL.SDL_SetHintWithPriority(name, value, SDL.SDL_HintPriority.SDL_HINT_OVERRIDE);
		}

		#endregion

```

flibitijibibo (commented):
We'll need to replace all of the remaining SetEnvironmentVariable calls, other than that just a couple quick review bits and this is good to go.

PR #441, opened 2023-10-03
Make GraphicsResource finalizer free resources... eventually

kg:
This PR refactors disposal of GraphicsResources:
* FNA3D handle disposal is performed in a centralized place (GraphicsResourceHandles)
* All GraphicsResources have a new virtual method that creates a GraphicsResourceHandles that can be used to dispose their handles
* The GraphicsResource finalizer will, if it was never disposed, grab the handles and safely push them into a queue on the GraphicsDevice
* GraphicsDevice.Present and GraphicsDevice.Dispose both empty the queue and dispose any handles that are in it, cleaning up leaked objects

### DIFF
--- a/src/Graphics/GraphicsResourceHandles.cs
+++ b/src/Graphics/GraphicsResourceHandles.cs
@@ -3,8 +3,9 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 
-namespace Microsoft.Xna.Framework.Graphics {
-    internal struct GraphicsResourceHandles
+namespace Microsoft.Xna.Framework.Graphics
+{
+	internal struct GraphicsResourceHandles
 	{
 		public IntPtr effect;
 		public IntPtr indexBuffer;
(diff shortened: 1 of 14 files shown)


kg:
Not sure what the hell visual studio did to the whitespace here. The editorconfig file didn't work I guess... Will fix it manually

src/Graphics/GraphicsResourceHandles.cs
```diff
@@ -0,0 +1,91 @@
...
+				);
+			}
+
+			if (mallocedPointer != IntPtr.Zero)
+			{
+				FNAPlatform.Free(mallocedPointer);
+			}
+
+			// Zero out all our fields to prevent a double-free
+			this = default;
```
kg, line 88: I just discovered this won't build in older versions of VS, it needs to be default(GraphicsResourceHandles)

flibitijibibo:
Superceded by #442

PR #442, opened 2023-10-04
Emergency dispose graphics resources from finalizer

thatcosmonaut:
Alternate implementation of #441 

Note that this definitely allocates in the "sad path" where the client loses track of a resource and the finalizer is invoked. They shouldn't do that and we do warn them, so I don't think it's a huge deal.

This also moves the Effect state changes pointer to the GraphicsDevice, which simplifies cleanup.

### DIFF
--- a/src/Graphics/Effect/Effect.cs
+++ b/src/Graphics/Effect/Effect.cs
@@ -307,6 +307,22 @@ namespace Microsoft.Xna.Framework.Graphics
 
 		#endregion
 
+		#region Emergency Disposal
+
+		internal override GraphicsResourceDisposalHandle[] CreateDisposalHandles()
+		{
+			return new GraphicsResourceDisposalHandle[]
+			{
+				new GraphicsResourceDisposalHandle
+				{
+					disposeAction = FNA3D.FNA3D_AddDisposeEffect,
+					resourceHandle = glEffect
+				}
+			};
+		}
+
+		#endregion
+
 		#region Protected Methods
 
 		protected override void Dispose(bool disposing)
--- a/src/Graphics/GraphicsDevice.cs
+++ b/src/Graphics/GraphicsDevice.cs
@@ -9,6 +9,7 @@
 
 #region Using Statements
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Runtime.InteropServices;
 #endregion
@@ -486,6 +487,8 @@ namespace Microsoft.Xna.Framework.Graphics
 						Disposing(this, EventArgs.Empty);
 					}
 
+					FlushEmergencyDisposalQueue();
+
 					/* Dispose of all remaining graphics resources before
 					 * disposing of the GraphicsDevice.
 					 */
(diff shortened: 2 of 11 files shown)


src/Graphics/RenderTarget2D.cs
```diff
@@ -208,5 +208,49 @@ protected internal override void GraphicsDeviceResetting()
...
+			{
+				handles[index] = new GraphicsResourceDisposalHandle
+				{
+					disposeAction = FNA3D.FNA3D_AddDisposeRenderbuffer,
+					resourceHandle = glDepthStencilBuffer
+				};
+				index += 1;
+			}
+			
+			return handles;
```
kg, line 251: Don't we also need AddDisposeTexture? You're not calling base.CreateDisposalHandles and carrying it over

kg:
I don't love the duplication, but it's very low risk since it's not like Texture.Dispose is ever going to change, right?

src/Graphics/RenderTarget2D.cs
thatcosmonaut: That's right, nice catch

src/Graphics/Effect/Effect.cs
```diff
@@ -73,8 +73,6 @@ public int GetHashCode(IntPtr obj)
 			}
 		}
 
-		private IntPtr stateChangesPtr;
```
flibitijibibo, line 76: Let's commit this part separately, this on its own lgtm

src/Graphics/GraphicsDevice.cs
```diff
@@ -547,6 +571,30 @@ internal void RemoveResourceReference(WeakReference resourceReference)
 
 		#endregion
 
+		#region Emergency Disposal / Finalization
+
+		ConcurrentQueue<GraphicsResourceDisposalHandle> emergencyDisposalQueue = new ConcurrentQueue<GraphicsResourceDisposalHandle>();
```
flibitijibibo, line 576: We can put this declaration under Private Disposal Variables

src/Graphics/GraphicsResource.cs
```diff
@@ -154,6 +161,12 @@ internal protected virtual bool IsHarmlessToLeakInstance
 			}
 		}
 
+		// This has to return an array because some resources have multiple handles... d'oh!
+		internal virtual GraphicsResourceDisposalHandle[] CreateDisposalHandles() 
+		{
+			return null;
```
flibitijibibo, line 167: I thought this could be a throw to make sure all types implement this, but I guess states and other types would just return null anyway... could probably throw in 
```suggestion
			// Only certain GraphicsResource types have pointers to dispose
			return null;
```

src/Graphics/GraphicsResourceDisposalHandle.cs
flibitijibibo, line 1: lgtm, just needs the license header

src/Graphics/GraphicsResource.cs
```diff
@@ -107,7 +107,14 @@ internal GraphicsResource()
 			}
 #endif
 
-			// FIXME: We really should call Dispose() here! -flibit
+			// While we only log in debug builds, in both debug and release builds we want to free
+			// any native resources associated with this object at the earliest opportunity.
+			// This will at least prevent you from running out of memory rapidly.
+			var handles = CreateDisposalHandles();
```
flibitijibibo, line 113:
```suggestion
			GraphicsResourceDisposalHandle[] handles = CreateDisposalHandles();
```

flibitijibibo (commented):
Just a few little things here and there - can merge this right after the tag, so we can let this cook for the whole month.

PR #445, opened 2023-10-14
Expose block size APIs

kg:


flibitijibibo (approved):
This looks okay, but it would be good to have documentation with an example like we have for the other extensions.

kg:
Where should the documentation go?

flibitijibibo:
I can copy markdown written here to wiki page for extensions. (Really wish wikis had better collaboration tools than "open or closed")

kg:
The `Texture.GetBlockSizeSquaredEXT` method allows querying the number of pixels in a block for compressed textures. You can use this in order to round sizes up/down to the nearest block, or to calculate how large a buffer needs to be for a compressed texture of a given size. Most compressed texture formats have a block size of `4` (and thus a `BlockSizeSquared` of `16`), but not all do. For formats that are not compressed, this method returns `1`.

The `Texture.GetFormatSizeEXT` method allows querying the number of bytes in a single block (or texel, for formats where the block size is `1`). For the `Color` format, for example, this is `4`.

By combining these two methods you can calculate the appropriate buffer size for any arbitrary GetData call, like so:
```csharp
var elementSize = Texture.GetFormatSizeEXT(texture.Format);
var blockSizeSquared = Texture.GetBlockSizeSquaredEXT(texture.Format);
var buffer = new byte[texture.Width * texture.Height * elementSize / blockSizeSquared];
texture.GetData(buffer);
```

flibitijibibo:
Added to the wiki, thanks!

PR #451, opened 2023-11-11
Flow Texture.Name through to FNA3D_SetTextureName

kg:


src/Graphics/Texture.cs
```diff
@@ -30,6 +30,29 @@ public int LevelCount
...
 
+		public override string Name
+		{
+			get
+			{
+				return base.Name;
+			}
+			set
+			{
+				// HACK: Avoid calling SetTextureName when the value hasn't changed.
```
flibitijibibo, line 41: This seems okay to me, it's basically just state shadowing...

src/Graphics/Texture.cs
```diff
@@ -30,6 +30,29 @@ public int LevelCount
...
+				if (value == base.Name)
+					return;
+
+				base.Name = value;
+
+				// HACK: Never pass a null string pointer through to SetTextureName.
+				// Since base.Name will be null by default, this will only happen if
+				//  you first set a name for the texture, then try to null it out.
+				if (value == null)
+					value = "";
```
flibitijibibo, line 51:
```suggestion
				// Avoid calling SetTextureName when the value hasn't changed.
				if (value == base.Name)
				{
					return;
				}

				base.Name = value;

				// Never pass a null string pointer through to SetTextureName.
				// Since base.Name will be null by default, this will only happen if
				//  you first set a name for the texture, then try to null it out.
				if (value == null)
				{
					value = string.Empty;
				}
```

flibitijibibo (approved):
This is actually okay in my book, can always push the null check further down the stack if we really need to.

PR #462, opened 2023-12-07
Emergency dispose graphics resources without double free

### LINKED ISSUE #461: Do a second take of #442
Test program:

```c#

class TextureContainer
{
    Texture2D tex;

    public TextureContainer(GraphicsDevice gd)
    {
        tex = new Texture2D(gd, 1, 1);
    }

    ~TextureContainer()
    {
        tex.Dispose();
    }
}

class Program : Game
{
    static void Main(string[] args)
    {
        using (Program p = new Program())
        {
            new GraphicsDeviceManager(p);
            p.Run();
        }
    }

    protected override void Draw(GameTime gameTime)
    {
        new TextureContainer(GraphicsDevice);
        if (gameTime.TotalGameTime > TimeSpan.FromSeconds(10))
        {
            GC.Collect();
        }
        base.Draw(gameTime);
    }
}
```

thatcosmonaut:
Resolves #461

Two changes from the reverted patch:

- Move the disposal enqueue into the block where the warning is logged in the finalizer.
- Set a flag when the disposal is enqueued and check the flag in Dispose so the native dispose function doesn't get called twice if Dispose is called afterward.

It should be noted that calling Dispose from a finalizer is a very bad idea, but at least it shouldn't crash now!

### DIFF
--- a/FNA.Core.csproj
+++ b/FNA.Core.csproj
@@ -184,6 +184,7 @@
 		<Compile Include="src\Graphics\GraphicsDeviceStatus.cs" />
 		<Compile Include="src\Graphics\GraphicsProfile.cs" />
 		<Compile Include="src\Graphics\GraphicsResource.cs" />
+		<Compile Include="src\Graphics\GraphicsResourceDisposalHandle.cs" />
 		<Compile Include="src\Graphics\IGraphicsDeviceService.cs" />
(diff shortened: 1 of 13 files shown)


thatcosmonaut:
Superseded by https://github.com/FNA-XNA/FNA/commit/2f8052ff88d3ec3999fd463b5a33370f226eb216

PR #464, opened 2023-12-11
Fix iOS/tvOS support in FNADllMap

TheSpydog:
Mono's dllmap considers "osx" to be the platform name for macOS, iOS, and tvOS. This PR brings FNADllMap in line with that convention.

### DIFF
--- a/src/Utilities/FNADllMap.cs
+++ b/src/Utilities/FNADllMap.cs
@@ -39,7 +39,7 @@ namespace Microsoft.Xna.Framework
 			{
 				return "windows";
 			}
-			else if (OperatingSystem.IsMacOS())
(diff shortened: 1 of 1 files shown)


kiddkaffeine:
Unfortunately, .net 7/8 appears to be stripping symbols even when you ask it not to.  The options to preserve individual symbols does still work, although there are quite a lot of them required for this and that list can change between FNA versions.

Anyway, on the FNA end, there is still one additional change required.  In MapAndLoad before the call to NativeLibrary.Load, we need to add the following:

`#if NET7_0_OR_GREATER
            if (mappedName == "__Internal") {
                return NativeLibrary.GetMainProgramHandle();
            }
#endif`

PR #469, opened 2023-12-28
Use SDL_GetKeyboardState instead of event polling

thatcosmonaut:
~~We got a report from the Celeste modding/speedrunning community that FNA was causing input latency on keyboard relative to XNA. FNA currently processes keyboard input by processing all keyboard events in PollEvents and setting the keys in an internal array. But according to the report, XNA actually gets a new state every time Keyboard.GetState is called, not just once per frame. This patch brings the behavior in line with XNA, and it should address latency by allowing input polling to occur closer to when input is actually processed by the game.~~

We got a report from the Celeste modding/speedrunning community that FNA was causing input latency on keyboard relative to XNA when an IME was installed. This patch gets the keyboard state directly all at once instead of iterating over keydown events. This might address some scenarios where the keydown event is intercepted by something. It should also be very slightly more efficient to not have to call `keys.Contains` on all keypresses.

### DIFF
--- a/src/FNAPlatform/SDL2_FNAPlatform.cs
+++ b/src/FNAPlatform/SDL2_FNAPlatform.cs
@@ -45,6 +45,14 @@ namespace Microsoft.Xna.Framework
 
 		#endregion
 
+		#region Keyboard State
+
+		// Used for special copy-paste text input
+		private static bool LeftControlDown = false;
+		private static bool RightControlDown = false;
+
+		#endregion
+
 		#region Init/Exit Methods
 
 		public static string ProgramInit(LaunchParameters args)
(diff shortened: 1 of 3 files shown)


src/Input/Keyboard.cs
```diff
@@ -25,9 +25,21 @@ public static class Keyboard
 		/// Returns the current keyboard state.
 		/// </summary>
 		/// <returns>Current keyboard state.</returns>
-		public static KeyboardState GetState()
+		public unsafe static KeyboardState GetState()
 		{
-			return new KeyboardState(keys);
+			// Returns a pointer that does not need to be freed.
```
flibitijibibo, line 30: We should inline this into `FNAPlatform.GetKeyboardState(List<Keys>)`, so that each `i` doesn't have to call the delegate in a tight loop. This also makes it possible to move the ControlDown bools to SDL2_FNAPlatform.

src/FNAPlatform/SDL2_FNAPlatform.cs
```diff
@@ -954,53 +954,68 @@ public static void UnregisterGame(Game game)
...
+					if (key == Keys.LeftControl)
+					{
+						Keyboard.LeftControlDown = true;
+					}
+					else if (key == Keys.RightControl)
+					{
+						Keyboard.RightControlDown = true;
+					}
+
+					if (evt.key.repeat > 0)
```
flibitijibibo, line 966: Super tedious but if we can invert this to be `if (evt.key.repeat == 0)` it will help shrink the diff of this if block which is otherwise pretty much the same as it was before.

flibitijibibo (changes requested):
Functionally this makes sense, so we can go into the polishing stage at this point.

flibitijibibo (approved):
Took care of the notes, as long as this works as intended this is ready to go.

flibitijibibo:
One last thing: Does the issue go away if we move our existing keydown/keyup events to an event watch? It's probably the best way to get these events without intercepting WndProc directly.

7aGiven:
> May as well update this since it came up recently: We have reason to believe that XNA actually calls GetAsyncKeyState directly when GetState is called, and SDL doesn't expose a direct mapping to this (nor should it, woof). This _may_ improve behavior but accurate behavior is likely impossible without insane levels of intervention in this specific class...

XNA call GetKeyboardState when GetState is called. It return all key state one time. So it update at message loop, Don't worry SDL not implement it.

7aGiven:
It maybe useful that SDL_WINDOWS_RAW_KEYBOARD=1

PR #471, opened 2024-01-01
FNA: Add last possible orientation value PORTRAIT_FLIPPED

jfmu:
… or PortraitUpsideDown. This really crashes on tablets otherwise. Corresponding pull request is in FNA3D, though the DisplayOrientation enum there is not used at all yet. https://github.com/FNA-XNA/FNA3D/pull/190

### DIFF
--- a/src/DisplayOrientation.cs
+++ b/src/DisplayOrientation.cs
@@ -34,6 +34,10 @@ namespace Microsoft.Xna.Framework
 		/// <summary>
 		/// The display is rotated as portrait, where height is greater than width.
 		/// </summary>
-		Portrait = 4
+		Portrait = 4,
+		/// <summary>
+		/// The display is rotated as portrait, where height is greater than width, but upside down.
+		/// </summary>
+		PortraitUpsideDown = 8
 	}
 }
--- a/src/FNAPlatform/SDL2_FNAPlatform.cs
+++ b/src/FNAPlatform/SDL2_FNAPlatform.cs
@@ -873,6 +873,9 @@ namespace Microsoft.Xna.Framework
 				case SDL.SDL_DisplayOrientation.SDL_ORIENTATION_PORTRAIT:
 					return DisplayOrientation.Portrait;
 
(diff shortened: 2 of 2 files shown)


flibitijibibo:
I think this has to be marked as an extension first, but this might be for @TheSpydog, since I thought this was either documented or translated properly for iOS.

flibitijibibo:
Are we able to add a case to this and treat it as Portrait?

https://github.com/FNA-XNA/FNA/blob/master/src/FNAPlatform/SDL2_FNAPlatform.cs#L863

TheSpydog:
Thanks for the PR / report!

I don't think this is the right solution, because the actual issue is more sinister: Windows builds have no business handling display orientation logic in the first place. The fact that any of this is happening in the first place is a bug. The only platforms that should care about orientation are [iOS and Android](https://github.com/FNA-XNA/FNA/blob/a5920865ab28dcd9d27fca22e03f2658e804b07b/src/FNAPlatform/FNAWindow.cs#L158). We're missing some logic to prevent other platforms from reacting to these orientation events.

However, even if we fixed that bug, this crash will still happen on iPad if `UIInterfaceOrientationPortraitUpsideDown` is listed as a supported orientation in the Info.plist. So we should address that as well, probably via flibit's proposed solution above.

TheSpydog:
> However, even if we fixed that bug, this crash will still happen on iPad if `UIInterfaceOrientationPortraitUpsideDown` is listed as a supported orientation in the Info.plist. So we should address that as well, probably via flibit's proposed solution above.

Confirmed that this proposed fix works on iPad, so that change should definitely be made.

flibitijibibo:
Let's go ahead and add that line in then - will push it some time tomorrow if nobody beats me to this.

Thanks for catching this!

jfmu:
Cool..thanks :-)

flibitijibibo:
https://github.com/FNA-XNA/FNA/commit/303edb551565619f8100ad9b71fa7719b1a093f6

PR #483, opened 2024-03-21
Detect AOT more accurately

ryancheung:
For projects with `<PublishAot>true</PublishAot>`, `RuntimeFeature.IsDynamicCodeCompiled` return `false` in debug build. 
See https://github.com/dotnet/runtime/issues/94380#issuecomment-1794352389

### DIFF
--- a/src/Utilities/FNADllMap.cs
+++ b/src/Utilities/FNADllMap.cs
@@ -91,7 +91,9 @@ namespace Microsoft.Xna.Framework
 		[ModuleInitializer]
 		public static void Init()
 		{
-			if (!RuntimeFeature.IsDynamicCodeCompiled)
+  			var stackTrace = new StackTrace(false);
+			var isAot = stackTrace.GetFrame(0)?.GetMethod() is null;
+			if (isAot)
 			{
 				/* NativeAOT platforms don't perform dynamic loading,
 				 * so setting a DllImportResolver is unnecessary.


kg:
Do you know if this works right for mixed-mode mono AOT, where some frames could be running under the interpreter and others under AOT? Would matter for WASM, but not for any other targets, I think.

src/Utilities/FNADllMap.cs
```diff
@@ -91,7 +91,9 @@ private static string GetPlatformName()
 		[ModuleInitializer]
 		public static void Init()
 		{
-			if (!RuntimeFeature.IsDynamicCodeCompiled)
+  			var stackTrace = new StackTrace(false);
+			var isAot = stackTrace.GetFrame(0)?.GetMethod() is null;
+			if (isAot)
```
MichalStrehovsky, line 96: `StackFrame.GetMethod` will return non-null on native AOT if the method happened to be considered a target of reflection by the compiler at compile time. The api can and does return non-null with native AOT.

What problem is this code trying to solve? If setting the DllImportResolver is just unnecessary, would it be fine to just unconditionally set it?

ryancheung: How we can detect none-NativeAOT if had `<PublishAot>true</PublishAot>` set in .csproj ?

MichalStrehovsky: You can't. What problem is this code trying to solve?

flibitijibibo:
Basically what our DllMap does is disable itself for AOT, because for FNA one of two things will be true:

- All binaries will be statically linked so everything will be __Internal anyway
- When linking the NativeAOT binary we're inlining the PInvokes and dynamically linking directly to the native libraries in question

So what this is trying to do is make sure that FNADllMap isn't getting more involved than it should be and risking unnecessary errors/resource use; Ryan's idea makes sense but it seems like .NET isn't making this easy to check. (Of course, if CoreCLR added DllMap we wouldn't have to think about this at all...)

Ryan: If you find a better fix for this feel free to re-open or open a new PR.

ryancheung:
OK, fine.

MichalStrehovsky:
> So what this is trying to do is make sure that FNADllMap isn't getting more involved than it should be and risking unnecessary errors/resource use

The callbacks will not be called if there's no p/invokes that need it. In general there's no harm in just registering them.

What you'd want here is an API to find out whether particular DllImport was bound statically. This would have to be some new reflection API on MethodInfo, for example.

This is trying to approximate an API that doesn't exist (is DllImport bound statically?) with an _approximation_ of another API that also doesn't exist (is this native AOT?).

`RuntimeFeature.IsDynamicCodeCompiled` returns false as soon as PublishAot is in the project file, even for F5 debug scenarios that have a JIT. Official docs guide people to place PublishAot in the project file if there's any intention to publish AOT later (nobody should be doing `dotnet publish -p:PublishAot=true`). PublishAot should be in the project file so that one doesn't get surprised by wildly different codepaths at runtime that are only reachable after publish. There are often codepaths that just do `if (!RuntimeFeature.IsDynamicCodeSupported) throw new NotSupportedException();`

PR #485, opened 2024-04-13
Enable Alpha8 render targets

kg:
I vaguely recall this working in XNA - I think escape goat 2 used it.
Right now it works for D3D11 and Vulkan with this change, but the behavior is different when sampling from the resulting RT. 

D3D11:
d3d11

Vulkan:
vulkan

Maybe an sRGB vs linear thing? Not sure.


kg:
The difference is in the way blending/rendering works for the RT; sampling is consistent. Disabling blending when drawing to it makes it work the same on both.

flibitijibibo:
Let's add this when gpu is ready, we should make sure it works there first.

PR #486, opened 2024-04-13
Add Texture2D.GetDataPointerEXT

kg:
Without this, it's not possible to cleanly queue up texture readback operations to execute later, since GetData<T> checks the size of T against the texture format, instead of checking the size of the buffer - so you can't read a Color texture into a byte[] even if it's big enough.

### DIFF
--- a/src/Graphics/Texture2D.cs
+++ b/src/Graphics/Texture2D.cs
@@ -298,11 +298,30 @@ namespace Microsoft.Xna.Framework.Graphics
 				);
 			}
 
+			int elementSizeInBytes = MarshalHelper.SizeOf<T>();
+			ValidateGetDataFormat(Format, elementSizeInBytes);
+
+			GCHandle handle = GCHandle.Alloc(data, GCHandleType.Pinned);
+			GetDataPointerEXT(
+				level,
+				rect,
+				handle.AddrOfPinnedObject() + (startIndex * elementSizeInBytes),
+				elementCount * elementSizeInBytes
+			);
+			handle.Free();
+		}
+
+		public void GetDataPointerEXT(
+			int level,
+			Rectangle? rect,
+			IntPtr data,
+			int dataLengthBytes
+		) {
 			int subX, subY, subW, subH;
 			if (rect == null)
 			{
-				subX = 0;
-				subY = 0;
+                subX = 0;
+                subY = 0;
 				subW = Width >> level;
 				subH = Height >> level;
 			}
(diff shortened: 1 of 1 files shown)


flibitijibibo:
I might be misremembering but I thought this worked like SetData where byte[] was okay... are we just missing some math checks that exist in SetData?

kg:
> I might be misremembering but I thought this worked like SetData where byte[] was okay... are we just missing some math checks that exist in SetData?

The
```csharp
	int elementSizeInBytes = MarshalHelper.SizeOf<T>();
	ValidateGetDataFormat(Format, elementSizeInBytes);
```
should reject an attempt to download a Color surface into a byte[], I think. IIRC the last time I tried it, it didn't work.

kg:
Alternate proposal: Don't do this, instead we add GetDataPointerAsyncEXT/SetDataPointerAsyncEXT, which only work on the SDL_GPU backend. If we had those I'd use them instead

kg:
Is there anything I could do to land this one?

flibitijibibo:
Meant to check this but never got around to it... I'm still _pretty_ sure this works; the format size of any SurfaceFormat % sizeof(byte) should always be 0, shouldn't it? A small test that just gets a blank 16x16 Color texture should be able to verify this.

kg:
It does appear to be working now, I wonder why I couldn't get it to work before... the generic instances for GetData<T> are still a little annoying to work around though, and it would be nice to be able to use a buffer outside of the managed heap. Will think it over a bit.

kg:
So the use case is this scenario, where I queue up readbacks to happen at the 'right time'. Is there an obvious solution for this I'm missing other than the awful cascade of elseifs and casts? You can see what this PR allows in the else block at the end.

```csharp
                if (rb.Destination is byte[] ba) {
                    rb.Source.GetData(0, null, ba, 0, ba.Length);
                } else if (rb.Destination is ushort[] ua) {
                    rb.Source.GetData(0, null, ua, 0, ua.Length);
                } else if (/* ... */) {
                    // ...
                } else {
                    var gch = GCHandle.Alloc(rb.Destination, GCHandleType.Pinned);
                    try {
                        var size = Marshal.SizeOf(rb.Destination.GetType().GetElementType()) * rb.Destination.Length;
                        rb.Source.GetDataPointerEXT(0, null, gch.AddrOfPinnedObject(), size);
                    } finally {
                        gch.Free();
                    }
                }
```

flibitijibibo:
I guess the weird part is why rb.Destination is ambiguous like that... is this a case of cascading templates or something? Are the templates super important in some way?

kg:
rb.Destination is an Array, since i just have a list of readbacks queued up. I guess I could create generic instances with virtual methods, and that's probably not bad.

src/Graphics/Texture2D.cs
```diff
@@ -298,11 +298,30 @@ int elementCount
...
+			IntPtr data,
+			int dataLengthBytes
+		) {
 			int subX, subY, subW, subH;
 			if (rect == null)
 			{
-				subX = 0;
-				subY = 0;
+                subX = 0;
+                subY = 0;
```
flibitijibibo, line 324:
```suggestion
				subX = 0;
				subY = 0;
```

flibitijibibo (approved):
Looked around the Graphics namespace and I _think_ this is the only qualifying class to need GetDataPointer since buffers never need it and 3D wouldn't support it anyway - I'd hate to see a renderer that needs readback on cubes.

Once the whitespace is fixed we'll go ahead with this one.

PR #487, opened 2024-06-03
Don't throw exception for missing Video until it's played

redmcg:
This is a PR I originally raised against the `wine-mono` fork to fix an issue with Hell Yeah! Wrath of the Dead Rabbit (205230):
https://github.com/madewokherd/FNA/pull/6

@madewokherd and I thought it would be worth checking to see if this was something wanted upstream.

The PR changes the behavior when a Video file is missing to match what Windows does. Windows will only throw an exception when you attempt to play the Video (rather than when loading the resource).

The aim of the patch is match Windows behavior, and therefore the focus is on the scenario where an XNB file is present. But as it is, this will also modify the behavior when:
- the XNB file is missing; and
- calls to `FromUriEXT`

which may not be desirable.

### DIFF
--- a/src/Content/ContentReaders/VideoReader.cs
+++ b/src/Content/ContentReaders/VideoReader.cs
@@ -41,10 +41,11 @@ namespace Microsoft.Xna.Framework.Content
 			/* The path string includes the ".wmv" extension. Let's see if this
 			 * file exists in a format we actually support...
 			 */
+			string origpath = path;
 			path = Normalize(path.Substring(0, path.Length - 4));
 			if (String.IsNullOrEmpty(path))
 			{
-				throw new ContentLoadException();
+				path = origpath;
 			}
 
 			int durationMS = input.ReadObject<int>();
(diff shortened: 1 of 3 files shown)


flibitijibibo:
Interesting, so it's kind of like Song where it's really just metadata that gets ignored until VideoPlayer uses it... works for me, but I may have another way of doing this - will report back in a moment.

flibitijibibo:
Did my own version of this - the main thing is that raw file behavior is preserved, which is what FNA games will prefer, but I redesigned the Video class in such a way that it acts more like Song, so VideoPlayer now does most of the work:

https://github.com/FNA-XNA/FNA/commit/e86dbccba022f5889a90f7e6ed197834650530bb

Thanks for the PR, this was very useful information to have!

flibitijibibo:
Actually woops, I forgot about the path normalizer - I'll apply this next.

flibitijibibo:
https://github.com/FNA-XNA/FNA/commit/65f98a8f99a01a3548ed0e01c3782e402cd2176a

redmcg:
Thanks @flibitijibibo , can confirm merging this in to wine-mono FNA fixes the issue. :+1:

PR #494, opened 2024-09-14
SDL3

thatcosmonaut:
Opening this up since the SDL3 bindgen project is far enough along. I've got the gamepad stuff converted already, will continue to port the rest.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,3 +10,6 @@
 [submodule "lib/Theorafile"]
 	path = lib/Theorafile
 	url = https://github.com/FNA-XNA/Theorafile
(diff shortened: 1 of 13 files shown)


flibitijibibo:
Remaining issues are tagged with `FIXME CSHARP` or `FIXME SDL3`. Other than that I only had to change the namespace declaration to use braces, .NET4 was happy with the rest!

flibitijibibo:
Everything is now hooked up and games are starting to boot with this - the remaining work should now be entirely in SDL3_FNAPlatform, unless we decide to wedge a window pointer into the text input system (probably will just use the current focused window, I dunno...)

PR #501, opened 2024-11-18
Removed graphics device reset when display event enum is not orientation.

### LINKED ISSUE #500: Game Window blurs completely when monitor is disconnected.

DoubleBullet:
- SDL2_FNAPlatform.cs: Removed graphics device reset when display event enum is not orientation.

This change no longer resets the viewport to 8x8 when the display is connected or disconnected for a full screen window.

The reason why the viewport is specifically reset to 8x8 is because the FNA3D.FNA3D_GetMaxMultiSampleCount returns 8.

Resetting might make sense if the monitor was swapped entirely with another one. I'm not sure how I can track for when that happens.

This has been tested on Windows, but not linux and mac.

Fixes #500

### DIFF
--- a/src/FNAPlatform/SDL2_FNAPlatform.cs
+++ b/src/FNAPlatform/SDL2_FNAPlatform.cs
@@ -1194,14 +1194,6 @@ namespace Microsoft.Xna.Framework
 							);
 						}
 					}
-					else
-					{
-						// Just reset, this is probably a hotplug
-						game.GraphicsDevice.Reset(
-							game.GraphicsDevice.PresentationParameters,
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Looking at the git history this was added because the GraphicsDevice.Adapter property becomes out-of-date when this happens. After this event happens, are you able to find the GraphicsDevice.Adapter instance in GraphicsAdapter.Adapters?

Either way we could probably just manually reassign the adapter without doing a whole reset... and for all I know the instance is outdated on XNA too, I'd have to add a second monitor to my VM somehow.

PR #502, opened 2024-11-18
Avoid doing full device resets on monitor hotplug events

### LINKED ISSUE #500: Game Window blurs completely when monitor is disconnected.

flibitijibibo:
Fixes #500

@DoubleBullet, does this also fix the issue?

### DIFF
--- a/src/FNAPlatform/SDL2_FNAPlatform.cs
+++ b/src/FNAPlatform/SDL2_FNAPlatform.cs
@@ -1196,9 +1196,8 @@ namespace Microsoft.Xna.Framework
 					}
 					else
 					{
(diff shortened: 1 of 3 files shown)


DoubleBullet:
This works great.
It fixes the issue.

I guess I'll close my PR and issue.
Thanks for the help!

DoubleBullet:
I need to add that we need to test this on Mac and Linux if possible.
I want to make sure that this is consistent on all platforms.
I can test linux in a bit. Though I don't have access to Mac.

flibitijibibo:
I'll test with AV1/2 today - this was the main thing we were intending to fix so as long as something weird hasn't popped up in the last 2 years this should be good to go!

PR #505, opened 2024-12-08
Gamepad fixes and error handling

kg:


src/FNAPlatform/SDL3_FNAPlatform.cs
```diff
@@ -2017,9 +2017,19 @@ private static void INTERNAL_AddInstance(uint dev)
 
 			// Open the device!
 			INTERNAL_devices[which] = SDL.SDL_OpenGamepad(dev);
+			string error = SDL.SDL_GetError();
+			if (!string.IsNullOrEmpty(error)) {
```
flibitijibibo, line 2021: Similar to the other checks we should use tue return value of the function instead of the error buffer

flibitijibibo (changes requested):
One note applied to various places in the fix

flibitijibibo:
Fast-tracked the fixes, so it's just the error logging that's left: https://github.com/FNA-XNA/FNA/commit/6415dfd6383242fa41ea56016a7c8f8f34a63ce1

PR #508, opened 2024-12-20
Add IsMinimizedEXT and IsMaximizedEXT

kg:


flibitijibibo:
We really only have IsBorderlessEXT because we inherited it from MonoGame, and it's not my favorite extension - definitely better to use the window handle directly and make use of SDL's window flags _and_ events (which can be tracked with AddEventWatch), which cover everything without needing anyone in the middle.

PR #509, opened 2024-12-25
Add ByteEXT and UShortEXT and enable as texture and RT formats

kg:
Pairs with https://github.com/FNA-XNA/FNA3D/pull/228


flibitijibibo (approved):
Just waiting on the FNA3D side first.

PR #517, opened 2025-03-24
Use code point on OnTextInput

Cn-mjt44:
In SDL2/3, text input event will only return a complete unicode character. If this part use char in csharp, It can't explane a complete unicode character at one times.

### DIFF
--- a/src/FNAPlatform/SDL2_FNAPlatform.cs
+++ b/src/FNAPlatform/SDL2_FNAPlatform.cs
@@ -947,6 +947,29 @@ namespace Microsoft.Xna.Framework
 			activeGames.Remove(game);
 		}
 
+
+		private static int UTF8ToUnicode(int utf8)
+		{
+			int
+				byte4 = utf8 & 0xFF,
+				byte3 = (utf8 >> 8) & 0xFF,
+				byte2 = (utf8 >> 16) & 0xFF,
+				byte1 = (utf8 >> 24) & 0xFF;
+
+			if (byte1 < 0x80)
+				return byte1;
+			else if (byte1 < 0xC0)
+				return -1;
+			else if (byte1 < 0xE0 && byte2 >= 0x80 && byte2 < 0xC0)
+				return (byte1 % 0x20) * 0x40 + (byte2 % 0x40);
+			else if (byte1 < 0xF0 && byte2 >= 0x80 && byte2 < 0xC0 && byte3 >= 0x80 && byte3 < 0xC0)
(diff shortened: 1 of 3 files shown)


flibitijibibo:
I might be mixing up types here but what we're intending to do with the current code is convert from UTF8 to wchar_t, which is what `char` _should_ be in C#. The text buffer in SDL3 got expanded to support whole strings, which is how we handle console text input. Is there a specific test we can try on the various platforms to see the before/after for this change?

Cn-mjt44:
@flibitijibibo No, I'm sorry, I don't have console. So I can't test for console.
But, even it push whole string to `SDL_TextInput`, it can still handel this. Because, it basicly turning the UTF-8 string to UTF-32 string. And then foreach every UTF-32 character push to OnTextInput. The UTF32 character is equle to unicode codepoint. So every character will be validity in unicode.

Cn-mjt44:
For example, `🤣` is `U+1F923`. If you try to repecent in c# string(utf-16 string), it will take 2 char. If you only takeing one utf-16 character from there, it will take no mean for text.

flibitijibibo:
Interesting - I wonder if this is a serious problem for platforms where sizeof(wchar_t) is 2 rather than 4.

The overall purpose of the patch makes sense to me but I'm surprised it requires so much manual intervention - should we isolate translation to a helper file maybe? Is this something upstream dotnet should care about?

flibitijibibo:
Understood - I've marked this as an extension bug for the moment; if it works locally great, but I assume this subject has a bunch of weird caveats so I'll let others chime in to make sure we get the next revision of TextInputEXT correct, since it's hard to modify extensions unless we're _really_ sure the change is both necessary and absolutely correct.

Expect some reviews over time, either way: Thanks for checking this out!

Cn-mjt44:
image

Cn-mjt44:
@flibitijibibo Can I add some code about clipbord in this?

flibitijibibo:
We're close to the end of FNA's Extension model, but we don't want to block extensions in progress, so I'm doing an approval system... if you get a lot of e-mails, apologies for that! Anyway, for this one:

I don't want to merge the new events since SDL_AddEventWatch does a good job without needing new interfaces, but we definitely want to figure out a solution for the Windows UTF-8 problem. How about this: Instead of changing the existing `Action<char>`, let's add a new `Action<byte*> OnTextInputUTF8` that just passes the pointer raw, and applications can do whatever decoding they need to do. I'll file a separate ticket for this so we don't lose track!

PR #518, opened 2025-03-31
Flush scissor state in SetRenderTargets

kiddkaffeine:


flibitijibibo (approved):
This should be okay, we may want to add _why_ we need to flush scissor state there in case it comes up years later or something.

kiddkaffeine:
Good suggestion; updated comment.

Issue #519, opened 2025-04-11
SoundEffectInstance.Play sets max frequency ratio to FAUDIO_DEFAULT_FREQ_RATIO instead of FAUDIO_MAX_FREQ_RATIO


jjohnsnaill:
This results in sounds not being able to play faster than 200% speed if the pitch clamps are removed. Speeds below 50% work correctly.

flibitijibibo:
Should be easy enough to patch up, as soon as it works on your end send it our way and we'll merge ASAP.

PR #520, opened 2025-04-11
Fix SoundEffectInstance.Play setting max frequency ratio to FAUDIO_DEFAULT_FREQ_RATIO instead of FAUDIO_MAX_FREQ_RATIO

Ebon1:
Resolves https://github.com/FNA-XNA/FNA/issues/519


flibitijibibo:
Looks good - this does increase memory allocations in FAudio by a lot though, so I may put this behind an environment variable to keep memory use optimal for the unmodified XNA spec.

flibitijibibo:
https://github.com/FNA-XNA/FNA/commit/65fc2902dc8ee4d7851e53fc7ff2836f78c6174f

Issue #524, opened 2025-05-05
SDL3: Fullscreen button broken


kiddkaffeine:
When using FNA on SDL3 with macOS, pressing the fullscreen button will cause the application to fullscreen and then immediately exit.  It appears that the fullscreen flag is set, and then FNA will exit fullscreen.

Unsure if other platforms are affected.  This was not an issue with SDL2.

PR #525, opened 2025-05-05
Add support for SDL3 fullscreen events

kiddkaffeine:
Appears to address the issue mentioned in https://github.com/FNA-XNA/FNA/issues/524 on macOS.

### DIFF
--- a/src/FNAPlatform/SDL3_FNAPlatform.cs
+++ b/src/FNAPlatform/SDL3_FNAPlatform.cs
@@ -1031,6 +1031,24 @@ namespace Microsoft.Xna.Framework
 					{
 						SDL.SDL_EnableScreenSaver();
 					}
+
+					// Full screen
+					else if (evt.type == (uint) SDL.SDL_EventType.SDL_EVENT_WINDOW_ENTER_FULLSCREEN)
+					{
+						Object manager = game.Services.GetService(typeof(IGraphicsDeviceManager));
+						if (manager is GraphicsDeviceManager)
+						{
+							((GraphicsDeviceManager) manager).IsFullScreen = true;
+						}
+					}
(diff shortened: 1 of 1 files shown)


src/FNAPlatform/SDL3_FNAPlatform.cs
```diff
@@ -1031,6 +1031,24 @@ ref bool textInputSuppress
...
+						Object manager = game.Services.GetService(typeof(IGraphicsDeviceManager));
+						if (manager is GraphicsDeviceManager)
+						{
+							((GraphicsDeviceManager) manager).IsFullScreen = true;
+						}
+					}
+					else if (evt.type == (uint) SDL.SDL_EventType.SDL_EVENT_WINDOW_LEAVE_FULLSCREEN)
+					{
+						Object manager = game.Services.GetService(typeof(IGraphicsDeviceManager));
+						if (manager is GraphicsDeviceManager)
```
flibitijibibo, line 1047: In case of weird timing issues:
```suggestion
						if (manager != null && manager is GraphicsDeviceManager)
						{
							((GraphicsDeviceManager) manager).IsFullScreen = true;
						}
					}
					else if (evt.type == (uint) SDL.SDL_EventType.SDL_EVENT_WINDOW_LEAVE_FULLSCREEN)
					{
						Object manager = game.Services.GetService(typeof(IGraphicsDeviceManager));
						if (manager != null && manager is GraphicsDeviceManager)
```

flibitijibibo (commented):
I tried to reproduce this with Wayland which has a similar user-facing option and couldn't make it happen, so aside from this tedious spec-lawyer fix this lgtm

src/FNAPlatform/SDL3_FNAPlatform.cs
kiddkaffeine: Null check added

PR #527, opened 2025-05-17
Optimize StateHash lookups

kg:
image
Using the default GenericEqualityComparer for these dictionaries adds overhead

### DIFF
--- a/src/Graphics/PipelineCache.cs
+++ b/src/Graphics/PipelineCache.cs
@@ -24,21 +24,23 @@ using System.Runtime.InteropServices;
 
 namespace Microsoft.Xna.Framework.Graphics
 {
-    #region Internal PSO Hash Struct
+	#region Internal PSO Hash Struct
 
-    internal sealed class StateHashComparer : IEqualityComparer<StateHash> {
+	internal sealed class StateHashComparer : IEqualityComparer<StateHash> {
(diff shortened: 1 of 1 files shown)


flibitijibibo (approved):
Looks like it just needs whitespace cleanup!

kg:
> Looks like it just needs whitespace cleanup!

The whitespace looks perfect on my end, any idea what the problem is? I'm stumped

thatcosmonaut:
Looks to me like a tab got converted to spaces up there.

src/Graphics/PipelineCache.cs
```diff
@@ -26,6 +26,20 @@ namespace Microsoft.Xna.Framework.Graphics
 {
 	#region Internal PSO Hash Struct
 
+	internal sealed class StateHashComparer : IEqualityComparer<StateHash> {
```
flibitijibibo, line 29:
```suggestion
	internal sealed class StateHashComparer : IEqualityComparer<StateHash>
	{
```

flibitijibibo: Thought I could commit this change via GitHub but for some reason it won't show the UI... this is the only thing left.

kg: Sorry for missing that one!

PR #529, opened 2025-06-02
Make AddTypeCreator func and some ContentTypeReader related classes public for AOT

Aranda:
**Motivation:**
On AOT platforms, we need to manually add these type creator functions for all of our content type readers so that they can be found (without reflection) when loading the binary-serialized XML content. I'm hoping it's not controversial to make these public so we don't have to maintain custom branch when updating FNA. Thanks!


flibitijibibo:
Aren't these covered by adding them to rd.xml?

PR #564, opened 2026-01-12
AV1 support [draft]

kiddkaffeine:
Currently you will be to import the branch of dav1dfile from this PR:
https://github.com/MoonsideGames/dav1dfile/pull/12

We may want to specify FPS in some other way.

There still seems to be a slight desync between equivalent Theora and AV1 video.


flibitijibibo (approved):
Everything looks good now - anything other than framerate sync issues left?

PR #569, opened 2026-01-13
Add optional FPS to FromUriEXT

kiddkaffeine:
This was mentioned as a possibility in chat.

Value is currently by Theora player.


flibitijibibo:
Going to see if we can add default arguments to an existing function and have the runtime keep compatibility, if not I'll probably skip this unless we start seeing a rise in video codecs that don't provide this by default.

Issue #576, opened 2026-04-14
PrepareType Behavior not the same


7aGiven:
```C#
		/// <summary>
		/// Removes Version, Culture and PublicKeyToken from a type string.
		/// </summary>
		/// <remarks>
		/// Supports multiple generic types (e.g. Dictionary&lt;TKey,TValue&gt;)
		/// and nested generic types (e.g. List&lt;List&lt;int&gt;&gt;).
		/// </remarks>
		/// <param name="type">
		/// A <see cref="System.String"/>
		/// </param>
		/// <returns>
		/// A <see cref="System.String"/>
		/// </returns>
		internal static string PrepareType(string type)
```
Exist a custom reader extends Microsoft.Xna.Framework.Content.ContentTypeReader;

The reader exist in dll and load it from AppDomain.CurrentDomain.AssemblyResolve;

The function remove the Version and Culture, It cause reload dll and throw InvalidCastException when cast type for ContentTypeReader result.

XNA not remove the Version and Culture, So it no error.

Note: when exist Culture and Version, It load successfully. not need PublicKeyToken.

For test code below.

```C#
    internal class Program
    {
        internal static void LaunchGame()
        {
            // Write some code to refence MyAssembly here.

            string q = "MyNamespace.MyClass, MyAssembly";

            Console.WriteLine("GetType Enter 1");
            Type type = Type.GetType(q);
            Console.WriteLine("GetType Leave 1");
            Console.WriteLine(type == null);
            Console.WriteLine(type);

            Console.WriteLine("GetType Enter 2");
            type = Type.GetType(q);
            Console.WriteLine("GetType Leave 2");
            Console.WriteLine(type == null);
            Console.WriteLine(type);

        }
    }

    public static class WindowsLaunch
    {
        [STAThread]
        private static void Main(string[] args)
        {
            AppDomain.CurrentDomain.AssemblyResolve += delegate (object sender, ResolveEventArgs sargs)
            {
                Console.WriteLine("AssemblyResolve|" + sargs.Name);
                if (!sargs.Name.Contains("MyAssembly"))
                    return null;
                return Assembly.Load(File.ReadAllBytes("MyAssemblyFile.dll"));
            };
            Program.LaunchGame();
        }
    }
```

~~SNIP~~

flibitijibibo:
Updated the OP to remove the XNA decomp, we can't look at this for clean-room reasons.

[...]

PR #577, opened 2026-04-15
fix PrepareType

7aGiven:
#576

### DIFF
--- a/src/Content/ContentTypeReaderManager.cs
+++ b/src/Content/ContentTypeReaderManager.cs
@@ -37,7 +37,7 @@ namespace Microsoft.Xna.Framework.Content
 
 		private static readonly Dictionary<Type, ContentTypeReader> contentReadersCache;
 
-		private static readonly string[] nestedMark = { "[[" };
+		private static readonly Regex regex = new Regex(@", (Microsoft.Xna.Framework.Graphics|Microsoft.Xna.Framework.Video|Microsoft.Xna.Framework|MonoGame.Framework), Version=.+?, Culture=.+?, PublicKeyToken=[^\]]+", RegexOptions.Compiled);
 
 		// Trick to prevent the linker removing the code, but not actually execute the code
 		private static bool falseflag = false;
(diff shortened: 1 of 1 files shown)


flibitijibibo (approved):
Feedback was good, but I haven't been able to test performance so I may make this the first commit after tagging 26.05.

Benchmarks from others' games are welcome; we expect there to be _some_ difference when loading a lot of XNB-packed content but given that we're calling Replace a whole lot less often this should hopefully be an _improvement_...?

7aGiven:
Is there some test cases for XNB performance?

Or you mean performance in real game?

flibitijibibo:
I'm planning to test with real games, but was also pointed to this as a possible test:

https://github.com/nkast/KniBenchmarks/tree/main/src/ContentBenchmarks

7aGiven:
For direct calls to this function 70,000 times, the time taken before the modification was 600ms, and after the modification it took 90ms. For KniBenchmarks, the number of calls did not reach that number, so the difference was not significant.

flibitijibibo:
Also, upon closer inspection, we do a pretty good job of caching what we get from the prepared types, so this shouldn't be hit more than a few times (if at all) anyway - so this should be good to go.

7aGiven:
I disagree with the new commit's `private static readonly string regexReplacement;`

Please don't add regexReplacement field.

For docs: https://learn.microsoft.com/en-us/dotnet/standard/base-types/grouping-constructs-in-regular-expressions#zero-width-positive-lookbehind-assertions

```C#
        private static readonly string assemblyName = "FNA, Version=26.3.0.0, Culture=neutral, PublicKeyToken=null";
        private static readonly Regex regex = new Regex(@"(?<=, )(Microsoft.Xna.Framework.Graphics|Microsoft.Xna.Framework.Video|Microsoft.Xna.Framework|MonoGame.Framework), Version=.+?, Culture=.+?, PublicKeyToken=[^\]]+", RegexOptions.Compiled);
        internal static string PrepareType(string type)
        {
            return regex.Replace(type, assemblyName);
        }
```

flibitijibibo:
I think I may have misread something; the latest commit just moves the value generated by string.Format and moves it to a static variable. Did it change behavior in some way?

7aGiven:
> I think I may have misread something; the latest commit just moves the value generated by string.Format and moves it to a static variable. Did it change behavior in some way?

I now believe there's nothing wrong with your changes. Therefore, please disregard my comment.

7aGiven:
I am sorry for misunderstood your commit just now; your commit is very good. The commit reduced the running time by half.

PR #583, opened 2026-06-02
Change SDL_EVENT_WINDOW_MOVED to SDL_EVENT_WINDOW_DISPLAY_CHANGED

7aGiven:
A trivial code modification, no for any bug

I'm not sure why SDL_EVENT_WINDOW_MOVED here.


flibitijibibo:
Historical artifact probably - I don't think we had that event type when I first wrote this.

PR #590, opened 2026-06-14
Don't center window after UserMovedResized on Windows

### LINKED ISSUE #580: GraphicsDeviceManager.ApplyChanges behavior not the same
For XNA: normal
For FNA: when resize window by mouse, it will be auto center.

Test case:
```C#
using System;

namespace ConsoleApp2
{
    public static class Program
    {
        [STAThread]
        {
            using (FNAGame g = new FNAGame())
                g.Run();
        }
    }

    class FNAGame : Game
    {
        GraphicsDeviceManager gdm;
        internal FNAGame()
        {
            gdm = new GraphicsDeviceManager(this);
            Window.AllowUserResizing = true;
        }
        protected override void Draw(GameTime gameTime)
        {
            base.Draw(gameTime);
            if (gdm.PreferredBackBufferWidth != gdm.GraphicsDevice.Viewport.Width || gdm.PreferredBackBufferHeight != gdm.GraphicsDevice.Viewport.Height)
            {
                gdm.PreferredBackBufferWidth = GraphicsDevice.Viewport.Width - 1;
                gdm.PreferredBackBufferHeight = GraphicsDevice.Viewport.Height - 1;
                gdm.ApplyChanges();
            }
        }
    }
}
```

7aGiven:
fix #580

### DIFF
--- a/src/FNAPlatform/SDL3_FNAPlatform.cs
+++ b/src/FNAPlatform/SDL3_FNAPlatform.cs
@@ -2839,8 +2839,7 @@ namespace Microsoft.Xna.Framework
 						{
 							int w, h;
 							SDL.SDL_Rect fullscreen;
-							uint displayID = SDL.SDL_GetDisplayForWindow(game.Window.Handle);
-							SDL.SDL_GetDisplayBounds(displayID, out fullscreen);
+							SDL.SDL_GetDisplayBounds(SDL.SDL_GetDisplayForWindow(game.Window.Handle), out fullscreen);
 							SDL.SDL_GetWindowSize(game.Window.Handle, out w, out h);
 							if (w == fullscreen.w && h == fullscreen.h)
 							{
--- a/src/FNAPlatform/SDL3_FNAPlatform.cs
+++ b/src/FNAPlatform/SDL3_FNAPlatform.cs
@@ -2839,7 +2839,10 @@ namespace Microsoft.Xna.Framework
 						SDL.SDL_Rect fullscreen;
 						SDL.SDL_GetDisplayBounds(SDL.SDL_GetDisplayForWindow(game.Window.Handle), out fullscreen);
 						SDL.SDL_GetWindowSize(game.Window.Handle, out w, out h);
-						SDL.SDL_SetWindowPosition(game.Window.Handle, fullscreen.x, fullscreen.y);
+						if (w == fullscreen.w && h == fullscreen.h)
+						{
+							SDL.SDL_SetWindowPosition(game.Window.Handle, fullscreen.x, fullscreen.y);
+						}
 					}
 				}
 			}
--- a/src/FNAPlatform/SDL3_FNAPlatform.cs
+++ b/src/FNAPlatform/SDL3_FNAPlatform.cs
@@ -902,6 +902,7 @@ namespace Microsoft.Xna.Framework
 			char* charsBuffer = stackalloc char[32]; // SDL_TEXTINPUTEVENT_TEXT_SIZE
 			while (SDL.SDL_PollEvent(out evt))
 			{
+				game.Window.CurrentlyModalLoop = false;
 				// Keyboard
 				if (evt.type == (uint) SDL.SDL_EventType.SDL_EVENT_KEY_DOWN)
 				{
@@ -2817,9 +2818,23 @@ namespace Microsoft.Xna.Framework
 						game.RedrawWindow();
 						if (evt->window.data1 == 1)
 						{
-							game.Window.UserMovedResized = true;
+							game.Window.CurrentlyModalLoop = true;
 						}
 						return false;
+					}
+				}
+			}
+			else if (evt->type == (uint) SDL.SDL_EventType.SDL_EVENT_WINDOW_MOVED || evt->type == (uint) SDL.SDL_EventType.SDL_EVENT_WINDOW_RESIZED)
+			{
+				foreach (Game game in activeGames)
+				{
+					if (game.Window != null &&
+						evt->window.windowID == SDL.SDL_GetWindowID(game.Window.Handle))
+					{
+						if (game.Window.CurrentlyModalLoop)
+						{
+							game.Window.UserMovedResized = true;
+						}
 					}
 				}
 			}
--- a/src/FNAPlatform/SDL3_FNAPlatform.cs
+++ b/src/FNAPlatform/SDL3_FNAPlatform.cs
@@ -902,7 +902,7 @@ namespace Microsoft.Xna.Framework
 			char* charsBuffer = stackalloc char[32]; // SDL_TEXTINPUTEVENT_TEXT_SIZE
 			while (SDL.SDL_PollEvent(out evt))
 			{
-				game.Window.CurrentlyModalLoop = false;
+				game.Window.CurrentlyModalLoop = false; // Modal loop blocks SDL_PollEvent. Modal loop end when SDL_PollEvent return.
 				// Keyboard
 				if (evt.type == (uint) SDL.SDL_EventType.SDL_EVENT_KEY_DOWN)
 				{
@@ -2815,11 +2815,11 @@ namespace Microsoft.Xna.Framework
 					if (	game.Window != null &&
 						evt->window.windowID == SDL.SDL_GetWindowID(game.Window.Handle)	)
 					{
-						game.RedrawWindow();
-						if (evt->window.data1 == 1)
+						if (evt->window.data1 == 1) // Scheduled send during modal loop
 						{
 							game.Window.CurrentlyModalLoop = true;
 						}
+						game.RedrawWindow();
 						return false;
 					}
 				}
--- a/src/FNAPlatform/SDL3_FNAPlatform.cs
+++ b/src/FNAPlatform/SDL3_FNAPlatform.cs
@@ -2849,7 +2849,7 @@ namespace Microsoft.Xna.Framework
 							}
 							else if (game.Window.FixBorderlessFullscreen)
 							{
-								SDL.SDL_SetWindowPosition(game.Window.Handle, 0x2FFF0000 /*SDL_WINDOWPOS_CENTERED_MASK*/ | (int) displayID, 0x2FFF0000 /*SDL_WINDOWPOS_CENTERED_MASK*/ | (int) displayID);
+								SDL.SDL_SetWindowPosition(game.Window.Handle, fullscreen.x + (fullscreen.w - w) / 2, fullscreen.y + (fullscreen.h - h) / 2);
 								game.Window.FixBorderlessFullscreen = false;
 							}
 						}
--- a/src/FNAPlatform/SDL3_FNAPlatform.cs
+++ b/src/FNAPlatform/SDL3_FNAPlatform.cs
@@ -474,7 +474,8 @@ namespace Microsoft.Xna.Framework
 			int clientHeight,
 			bool wantsFullscreen,
 			string screenDeviceName,
-			ref string resultDeviceName
+			ref string resultDeviceName,
+			bool userMovedResized
 		) {
 			bool center = false;
 
@@ -508,7 +509,10 @@ namespace Microsoft.Xna.Framework
 				if (resize)
 				{
 					SDL.SDL_SetWindowSize(sdlWindow, clientWidth, clientHeight);
-					center = true;
+					if (!userMovedResized)
+					{
+						center = true;
+					}
 				}
 			}
 
@@ -2811,6 +2815,10 @@ namespace Microsoft.Xna.Framework
 						evt->window.windowID == SDL.SDL_GetWindowID(game.Window.Handle)	)
 					{
 						game.RedrawWindow();
+						if (evt->window.data1 == 1)
+						{
+							game.Window.UserMovedResized = true;
+						}
 						return false;
 					}
 				}
--- a/src/FNAPlatform/SDL3_FNAPlatform.cs
+++ b/src/FNAPlatform/SDL3_FNAPlatform.cs
@@ -2835,6 +2835,11 @@ namespace Microsoft.Xna.Framework
 						{
 							game.Window.UserMovedResized = true;
 						}
+						int w, h;
+						SDL.SDL_Rect fullscreen;
+						SDL.SDL_GetDisplayBounds(SDL.SDL_GetDisplayForWindow(game.Window.Handle), out fullscreen);
+						SDL.SDL_GetWindowSize(game.Window.Handle, out w, out h);
+						SDL.SDL_SetWindowPosition(game.Window.Handle, fullscreen.x, fullscreen.y);
 					}
 				}
 			}
--- a/src/FNAPlatform/SDL3_FNAPlatform.cs
+++ b/src/FNAPlatform/SDL3_FNAPlatform.cs
@@ -2835,13 +2835,23 @@ namespace Microsoft.Xna.Framework
 						{
 							game.Window.UserMovedResized = true;
 						}
-						int w, h;
-						SDL.SDL_Rect fullscreen;
-						SDL.SDL_GetDisplayBounds(SDL.SDL_GetDisplayForWindow(game.Window.Handle), out fullscreen);
-						SDL.SDL_GetWindowSize(game.Window.Handle, out w, out h);
-						if (w == fullscreen.w && h == fullscreen.h)
+						if (evt->type == (uint) SDL.SDL_EventType.SDL_EVENT_WINDOW_RESIZED && !SDL.SDL_GetWindowFlags(game.Window.Handle).HasFlag(SDL.SDL_WindowFlags.SDL_WINDOW_FULLSCREEN))
 						{
-							SDL.SDL_SetWindowPosition(game.Window.Handle, fullscreen.x, fullscreen.y);
+							int w, h;
+							SDL.SDL_Rect fullscreen;
+							uint displayID = SDL.SDL_GetDisplayForWindow(game.Window.Handle);
+							SDL.SDL_GetDisplayBounds(displayID, out fullscreen);
+							SDL.SDL_GetWindowSize(game.Window.Handle, out w, out h);
+							if (w == fullscreen.w && h == fullscreen.h)
+							{
+								SDL.SDL_SetWindowPosition(game.Window.Handle, fullscreen.x, fullscreen.y);
+								game.Window.FixBorderlessFullscreen = true;
+							}
+							else if (game.Window.FixBorderlessFullscreen)
+							{
+								SDL.SDL_SetWindowPosition(game.Window.Handle, 0x2FFF0000 | (int) displayID, 0x2FFF0000 | (int) displayID);
+								game.Window.FixBorderlessFullscreen = false;
+							}
 						}
 					}
 				}
--- a/src/FNAPlatform/SDL3_FNAPlatform.cs
+++ b/src/FNAPlatform/SDL3_FNAPlatform.cs
@@ -2835,7 +2835,7 @@ namespace Microsoft.Xna.Framework
 						{
 							game.Window.UserMovedResized = true;
 						}
-						if (evt->type == (uint) SDL.SDL_EventType.SDL_EVENT_WINDOW_RESIZED && !SDL.SDL_GetWindowFlags(game.Window.Handle).HasFlag(SDL.SDL_WindowFlags.SDL_WINDOW_FULLSCREEN))
+						if (evt->type == (uint) SDL.SDL_EventType.SDL_EVENT_WINDOW_RESIZED && (SDL.SDL_GetWindowFlags(game.Window.Handle) & SDL.SDL_WindowFlags.SDL_WINDOW_FULLSCREEN) == 0)
 						{
 							int w, h;
 							SDL.SDL_Rect fullscreen;
@@ -2849,7 +2849,7 @@ namespace Microsoft.Xna.Framework
 							}
 							else if (game.Window.FixBorderlessFullscreen)
 							{
-								SDL.SDL_SetWindowPosition(game.Window.Handle, 0x2FFF0000 | (int) displayID, 0x2FFF0000 | (int) displayID);
+								SDL.SDL_SetWindowPosition(game.Window.Handle, 0x2FFF0000 /*SDL_WINDOWPOS_CENTERED_MASK*/ | (int) displayID, 0x2FFF0000 /*SDL_WINDOWPOS_CENTERED_MASK*/ | (int) displayID);
 								game.Window.FixBorderlessFullscreen = false;
 							}
 						}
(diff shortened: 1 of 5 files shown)


flibitijibibo:
The variable seems kind of leaky, would it make more sense to have a bool for CurrentlyResizing that skips the size/position calls during a ClientSizeChanged callback?

7aGiven:
XNA just no longer center window after user resize window or user move window once.

I think I have not found a way to implement CurrentlyResizing. And may no need to implement CurrentlyResizing?

7aGiven:
The variable no leak. UserMovedResized mean User have moved window once or User have resized window once.

7aGiven:
@flibitijibibo The new commit fix that. SDL_WINDOWEVENT_EXPOSED(data1=1) is triggered while, click title bar but not real move or resize.

So I add variable CurrentlyModalLoop. It mean whether during modal loop.

Set UserMovedResized=true when SDL_EVENT_WINDOW_MOVED and SDL_EVENT_WINDOW_RESIZED during CurrentlyModalLoop=true.

But CurrentlyModalLoop not used for non Windows. It is only Windows feature. I'm not familiar with non-Windows event loops.

flibitijibibo:
There's an awful lot going on in here... is this game available somewhere that I can try this stuff out locally?

7aGiven:
> There's an awful lot going on in here... is this game available somewhere that I can try this stuff out locally?

I'm not sure what you mean. My understand is you want a test case. But it exist in #580?

Window center after user resize window in FNA. XNA not.
```C#
using System;

namespace ConsoleApp3
{
    public static class Program
    {
        [STAThread]
        {
            using (Game g = new FNAGame())
                g.Run();
        }
    }

    class FNAGame : Game
    {
        GraphicsDeviceManager gdm;
        internal FNAGame()
        {
            gdm = new GraphicsDeviceManager(this);
            Window.AllowUserResizing = true;
        }
        protected override void Draw(GameTime gameTime)
        {
            base.Draw(gameTime);
            if (gdm.PreferredBackBufferWidth != gdm.GraphicsDevice.Viewport.Width || gdm.PreferredBackBufferHeight != gdm.GraphicsDevice.Viewport.Height)
            {
                gdm.PreferredBackBufferWidth = GraphicsDevice.Viewport.Width - 1;
                gdm.PreferredBackBufferHeight = GraphicsDevice.Viewport.Height - 1;
                gdm.ApplyChanges();
            }
        }
    }
}
```

flibitijibibo:
Test cases are helpful, but a full shipping product is also helpful just to be sure we're not chasing after something extra unusual - I've not seen this amount of trouble from Windows resizing before so real-world data can be helpful for long-term regression testing.

flibitijibibo (approved):
Code looks okay but I'm going to have some people try this out first before merging, just in case.

src/GameWindow.cs
```diff
@@ -84,6 +84,14 @@ public virtual bool IsBorderlessEXT
 
 		#endregion
 
+		#region Internal Variables
+
+		internal bool CurrentlyModalLoop;
```
kg: It would be good to have a comment explaining what this represents, i.e. my guess:
On Windows when the user drags or resizes a window, a nested modal loop is started that consumes all mouse input until the mouse drag/resize operation ends. We need to track a guess about whether we are currently in a modal loop so we can detect user-generated drag/resize operations as distinct from ones caused by programmatic resize/movement. Am I correct?

Does this work for non-mouse-based movement like Aero Snap (win-left or win-right)?

src/FNAPlatform/SDL3_FNAPlatform.cs
```diff
@@ -2811,10 +2816,28 @@ private static unsafe bool Win32OnPaint(IntPtr userdata, SDL.SDL_Event* evt)
 						evt->window.windowID == SDL.SDL_GetWindowID(game.Window.Handle)	)
 					{
 						game.RedrawWindow();
+						if (evt->window.data1 == 1)
```
kg: From the SDL docs:
```
    SDL_EVENT_WINDOW_EXPOSED,           /**< Window has been exposed and should be redrawn, and can be redrawn directly from event watchers for this event.
                                             data1 is 1 for live-resize expose events, 0 otherwise. */
```
Might be good to have a comment here explaining this though because it took me a little digging to find this out.
Maybe you should set CurrentlyModalLoop before calling RedrawWindow in case the redraw code pumps the event loop?

src/GameWindow.cs
kg: Also, I don't see code anywhere to ever set CurrentlyModalLoop back to false. That could be a problem.

kg:
My only concern here other than the comments is that right now centering is important for fullscreen mode transitions. Without it, entering fullscreen mode could leave the game window hanging off the edge of the screen instead of placing it at `0, 0` where it belongs. And when you exit fullscreen mode with centering disabled, it's going to leave the game window at `0, 0`, which is kind of awkward.

I haven't had a chance to test this changeset locally and see how hard it would be to ensure it works for that, will try to find time later. It's possible the fullscreen mode transitions will magically work the right way because SDL is that good.

kg:
This PR does break fullscreen transitions:
image

I can confirm it works as intended otherwise though.

7aGiven:
> Test cases are helpful, but a full shipping product is also helpful just to be sure we're not chasing after something extra unusual - I've not seen this amount of trouble from Windows resizing before so real-world data can be helpful for long-term regression testing.

I understand. I pick the bug part from real game. It cause the bug.
```C#
using SDL3;
using System;
using System.Drawing;
using System.Runtime.InteropServices;
using System.Windows.Forms;

namespace ConsoleApp3
{
    public static class Program
    {
        [STAThread]
        {
            using (Game g = new FNAGame())
                g.Run();
        }
    }

    class FNAGame : Game
    {
        GraphicsDeviceManager gdm;
        int maxScreenW = 1920;
        int maxScreenH = 1200;
        int minScreenW = 800;
        int minScreenH = 600;
        int screenWidth;
        int screenHeight;
        internal FNAGame()
        {
            gdm = new GraphicsDeviceManager(this);
            Window.AllowUserResizing = true;
        }
        protected override void Draw(GameTime gameTime)
        {
            base.Draw(gameTime);
            bug(GraphicsDevice.Viewport.Width, GraphicsDevice.Viewport.Height);
        }
        private void bug(int width, int height)
        {
            Form form = (Form)Control.FromHandle(Window.Handle);
            form.MinimumSize = new Size(minScreenW, minScreenH);
            width = Math.Min(width, maxScreenW);
            height = Math.Min(height, maxScreenH);
            int viewportWidth = gdm.GraphicsDevice.Viewport.Width;
            int viewportHeight = gdm.GraphicsDevice.Viewport.Height;
            bool resized = (gdm.PreferredBackBufferWidth != gdm.GraphicsDevice.Viewport.Width || gdm.PreferredBackBufferHeight != gdm.GraphicsDevice.Viewport.Height);
            if (!resized)
            {
                if (form.ClientSize.Width < gdm.PreferredBackBufferWidth)
                {
                    width = form.ClientSize.Width;
                    resized = true;
                }
                if (form.ClientSize.Height < gdm.PreferredBackBufferHeight)
                {
                    height = form.ClientSize.Height;
                    resized = true;
                }
            }
            width &= 0x7FFFFFFE;
            height &= 0x7FFFFFFE;
            width = Math.Max(width, minScreenW);
            height = Math.Max(height, minScreenH);
            if (width != viewportWidth || height != viewportHeight || resized)
            {
                screenWidth = width;
                screenHeight = height;
                gdm.PreferredBackBufferWidth = screenWidth;
                gdm.PreferredBackBufferHeight = screenHeight;
                gdm.ApplyChanges();
                form.FormBorderStyle = FormBorderStyle.Sizable;
                form.SendToBack();
                form.BringToFront();
            }
            if (!gdm.SynchronizeWithVerticalRetrace)
            {
                gdm.SynchronizeWithVerticalRetrace = true;
                gdm.ApplyChanges();
            }
        }
    }

    class Control
    {
        private protected IntPtr window;
        private protected IntPtr handle;
        public virtual Size MinimumSize
        {
            set
            {
                SDL.SDL_GetWindowBordersSize(window, out int top, out int left, out int bottom, out int right);
                SDL.SDL_SetWindowMinimumSize(window, value.Width - left - right, value.Height - top - bottom);
            }
        }
        public void SendToBack()
        {
            SafeNativeMethods.SetWindowPos(new HandleRef(this.window, this.handle), NativeMethods.HWND_BOTTOM, 0, 0, 0, 0, 3);
        }
        public void BringToFront()
        {
            SafeNativeMethods.SetWindowPos(new HandleRef(this.window, this.handle), NativeMethods.HWND_TOP, 0, 0, 0, 0, 3);
        }
        public static Control FromHandle(IntPtr handle)
        {
            return new Form(handle);
        }
    }

    class Form:Control
    {
        public Form(IntPtr handle)
        {
            this.window = handle;
            this.handle = SDL.SDL_GetPointerProperty(SDL.SDL_GetWindowProperties(handle), SDL.SDL_PROP_WINDOW_WIN32_HWND_POINTER, IntPtr.Zero);
        }
        public FormBorderStyle FormBorderStyle
        {
            get
            {
                SDL.SDL_WindowFlags flags = SDL.SDL_GetWindowFlags(window);
                if ((flags & SDL.SDL_WindowFlags.SDL_WINDOW_RESIZABLE) != 0)
                    return FormBorderStyle.Sizable;
                if ((flags & SDL.SDL_WindowFlags.SDL_WINDOW_BORDERLESS) != 0)
                    return FormBorderStyle.None;
                throw new NotImplementedException();
            }
            set
            {
                if (value == FormBorderStyle.None)
                    SDL.SDL_SetWindowBordered(window, false);
                else if (value == FormBorderStyle.Sizable)
                    SDL.SDL_SetWindowBordered(window, true);
                else
                    throw new NotImplementedException();

            }
        }
        public Size ClientSize
        {
            get
            {
                SDL.SDL_GetWindowSize(window, out int w, out int h);
                return new Size(w, h);
            }
        }
    }
    internal static class SafeNativeMethods
    {
        [DllImport("user32.dll", CharSet = CharSet.Auto, ExactSpelling = true)]
        public static extern bool SetWindowPos(HandleRef hWnd, HandleRef hWndInsertAfter, int x, int y, int cx, int cy, int flags);
    }
    internal static class NativeMethods
    {
        public static HandleRef HWND_TOP = new HandleRef(null, (IntPtr)0);
        public static HandleRef HWND_BOTTOM = new HandleRef(null, (IntPtr)1);
    }
}
```

src/GameWindow.cs
7aGiven: set CurrentlyModalLoop=false when PollEvent() end. Modal loop block PollEvent until modal loop end.

During Modal Loop, event can received from SDL_EventFilter, event can't received from PollEvent.

SDL3 send SDL_EVENT_WINDOW_EXPOSED(data1=1) to keep draw during modal loop due to PollEvent blocked.

src/FNAPlatform/SDL3_FNAPlatform.cs
7aGiven: I agree all.

src/GameWindow.cs
7aGiven: This doesn't work for non-mouse-based movement like Aero Snap (win-left or win-right).

XNA is the same behavior for non-mouse-based movement. mean continue keep center window.

7aGiven:
> This PR does break fullscreen transitions: 
> 
> I can confirm it works as intended otherwise though.

Is it any way to reproduce the bug?

I write a test case but seem nothing happen?

Test case toggle Fullscreen when mouse click anywhere:
```C#
using Microsoft.Xna.Framework.Input;
using System;

namespace ConsoleApp3
{
    public static class Program
    {
        [STAThread]
        private static unsafe void Main(string[] args)
        {
            using (Game g = new FNAGame())
                g.Run();
        }
    }
    class FNAGame : Game
    {
        GraphicsDeviceManager gdm;
        internal FNAGame()
        {
            gdm = new GraphicsDeviceManager(this);
            Window.AllowUserResizing = true;
            IsMouseVisible = true;
        }
        bool click = false;
        protected override void Update(GameTime gameTime)
        {
            base.Update(gameTime);
            MouseState mouseState = Mouse.GetState();
            if (click)
            {
                if (mouseState.LeftButton == ButtonState.Released)
                    click = false;
            }
            else
            {
                if (mouseState.LeftButton == ButtonState.Pressed)
                {
                    click = true;
                    gdm.ToggleFullScreen();
                }
            }
        }
        protected override void Draw(GameTime gameTime)
        {
            base.Draw(gameTime);
            if (GraphicsDevice.PresentationParameters.BackBufferWidth != GraphicsDevice.Viewport.Width || GraphicsDevice.PresentationParameters.BackBufferHeight != GraphicsDevice.Viewport.Height)
            {
                gdm.PreferredBackBufferWidth = GraphicsDevice.Viewport.Width;
                gdm.PreferredBackBufferHeight = GraphicsDevice.Viewport.Height;
                gdm.ApplyChanges();
            }
            GraphicsDevice.Clear(Color.Black);
            using (SpriteBatch batch = new SpriteBatch(GraphicsDevice))
            {
                Texture2D texture = genTexture();
                batch.Begin(SpriteSortMode.Immediate, BlendState.AlphaBlend);
                batch.Draw(texture, Vector2.Zero, Color.White);
                batch.End();
                texture.Dispose();
            }
        }
        private Texture2D genTexture()
        {
            int width = GraphicsDevice.PresentationParameters.BackBufferWidth;
            int height = GraphicsDevice.PresentationParameters.BackBufferHeight;

            uint[] data = new uint[width * height];
            for (uint y = 0; y < height; y++)
                for (uint x = 0; x < width; x++)
                    data[y * width + x] = 3 * y + x;
            Texture2D texture = new Texture2D(GraphicsDevice, width, height);
            texture.SetData(data);
            return texture;
        }
    }
}
```

kg:
I'm using Window.IsBorderlessEXT not traditional fullscreen, sorry. I forgot to mention that.

src/GameWindow.cs
7aGiven: > It would be good to have a comment explaining what this represents, i.e. my guess: On Windows when the user drags or resizes a window, a nested modal loop is started that consumes all mouse input until the mouse drag/resize operation ends. We need to track a guess about whether we are currently in a modal loop so we can detect user-generated drag/resize operations as distinct from ones caused by programmatic resize/movement. Am I correct?

Your guess is correct.

7aGiven:
> I'm using Window.IsBorderlessEXT not traditional fullscreen, sorry. I forgot to mention that.

XNA set Borderless as below code:
```C#
        private void SetBorderless(bool isBordless)
        {
            Form form = (Form)Control.FromHandle(Window.Handle);
            form.FormBorderStyle = isBordless ? (Window.AllowUserResizing ? FormBorderStyle.Sizable : FormBorderStyle.FixedSingle) : FormBorderStyle.None;
        }
```
XNA seem not center window when enter or exit borderless mode too?

SDL_SetWindowFullscreen save location and size for windowed mode when enter fullscreen mode. set location and size when exit fullscreen mode.

7aGiven:
By the way. FNA IsFullScreen=true is borderless window fullscreen mode. FNA not implement exclusive fullscreen.

7aGiven:
> My only concern here other than the comments is that right now centering is important for fullscreen mode transitions. Without it, entering fullscreen mode could leave the game window hanging off the edge of the screen instead of placing it at `0, 0` where it belongs. And when you exit fullscreen mode with centering disabled, it's going to leave the game window at `0, 0`, which is kind of awkward.
> 
> I haven't had a chance to test this changeset locally and see how hard it would be to ensure it works for that, will try to find time later. It's possible the fullscreen mode transitions will magically work the right way because SDL is that good.

I think it shouldn't depend on center window.
For implement borderless fullscreen mode. There are two way:
```C
    // SDL_SetWindowFullscreen auto save and restore windowed position and size.
    // SDL_SetWindowFullscreenMode(window, NULL); // Paramter NULL mean default mode that borderless fullscreen mode
    SDL_SetWindowFullscreen(window, true);

    SDL_SetWindowFullscreen(window, false);
```
And
```C
    // save windowed position and size
    SDL_Rect windowed;
    bool oldflags = SDL_GetWindowFlags(window);
    if (!(oldflags & SDL_WINDOW_MAXIMIZED)) {
        SDL_GetWindowPosition(window, &windowed.x, &windowed.y);
        SDL_GetWindowSize(window, &windowed.w, &windowed.h);
    }

    // borderless fullscreen
    SDL_Rect fullscreen;
    SDL_DisplayID display = SDL_GetDisplayForWindow(window);
    SDL_GetDisplayBounds(display, &fullscreen);
    SDL_SetWindowBordered(window, false);
    SDL_SetWindowPosition(window, fullscreen.x, fullscreen.y);
    SDL_SetWindowSize(window, fullscreen.w, fullscreen.h);

    // restore windowed position and size
    if (!(oldflags & SDL_WINDOW_BORDERLESS)) {
        SDL_SetWindowBordered(window, true);
    }
    if (oldflags & SDL_WINDOW_MAXIMIZED) {
        SDL_MaximizeWindow(window);
    }
    else {
        SDL_SetWindowPosition(window, windowed.x, windowed.y);
        SDL_SetWindowSize(window, windowed.w, windowed.h);
    }
```

kg:
I don't have any preference for either of those proposed solutions, but right now the reality is that before this PR, IsBorderlessEXT worked for fullscreen and after this PR it doesn't because the centering is gone. I'm not sure what the best fix for that is. Ideally not many games are using it but I assume it was added based on popular demand... I think I'm using it because the old Vulkan backend used exclusive fullscreen mode, so I can probably safely remove it from anything new I release after today.

Maybe it's as simple as changing SetWindowBorderless in FNAPlatform to use SetWindowFullscreen instead in the case where the window size matches the current display size? But that feels a bit tricky to get right.

7aGiven:
I think when move window or resize window or set window borderless, check if borderless && window size is monitor size.

Then center window.

7aGiven:
> I don't have any preference for either of those proposed solutions, but right now the reality is that before this PR, IsBorderlessEXT worked for fullscreen and after this PR it doesn't because the centering is gone. I'm not sure what the best fix for that is. Ideally not many games are using it but I assume it was added based on popular demand... I think I'm using it because the old Vulkan backend used exclusive fullscreen mode, so I can probably safely remove it from anything new I release after today.
> 
> Maybe it's as simple as changing SetWindowBorderless in FNAPlatform to use SetWindowFullscreen instead in the case where the window size matches the current display size? But that feels a bit tricky to get right.

Thank you for your good idea. I did some commits to fix it.

7aGiven:
The PR maybe not a good idea due to I have not found a way to implement for Wayland.

7aGiven:
> The PR maybe not a good idea due to I have not found a way to implement for Wayland.

After going through the Wayland docs, it look like there is no API to set the window position or to center a window. So there is no this issue for Wayland.

src/FNAPlatform/SDL3_FNAPlatform.cs
```diff
@@ -2810,11 +2815,47 @@ private static unsafe bool Win32OnPaint(IntPtr userdata, SDL.SDL_Event* evt)
...
+							SDL.SDL_GetDisplayBounds(displayID, out fullscreen);
+							SDL.SDL_GetWindowSize(game.Window.Handle, out w, out h);
+							if (w == fullscreen.w && h == fullscreen.h)
+							{
+								SDL.SDL_SetWindowPosition(game.Window.Handle, fullscreen.x, fullscreen.y);
+								game.Window.FixBorderlessFullscreen = true;
+							}
+							else if (game.Window.FixBorderlessFullscreen)
+							{
+								SDL.SDL_SetWindowPosition(game.Window.Handle, 0x2FFF0000 | (int) displayID, 0x2FFF0000 | (int) displayID);
```
kg: Is this `SDL_WINDOWPOS_CENTERED_DISPLAY(displayID)`? A comment would be good.

src/FNAPlatform/SDL3_FNAPlatform.cs
```diff
@@ -2810,11 +2815,47 @@ private static unsafe bool Win32OnPaint(IntPtr userdata, SDL.SDL_Event* evt)
...
+				foreach (Game game in activeGames)
+				{
+					if (game.Window != null &&
+						evt->window.windowID == SDL.SDL_GetWindowID(game.Window.Handle))
+					{
+						if (game.Window.CurrentlyModalLoop)
+						{
+							game.Window.UserMovedResized = true;
+						}
+						if (evt->type == (uint) SDL.SDL_EventType.SDL_EVENT_WINDOW_RESIZED && !SDL.SDL_GetWindowFlags(game.Window.Handle).HasFlag(SDL.SDL_WindowFlags.SDL_WINDOW_FULLSCREEN))
```
kg: Nitpick (not super important):
Unfortunately, at least on NET4x (which FNA still supports), HasFlag accepts an `Enum`-typed argument, which I believe means it will be boxed. This means for every EVENT_WINDOW_RESIZED event processed a small amount of garbage will be allocated. If you look at the other code in this file the pattern typically used for this is to use `&` instead, i.e. for this case
```csharp
(SDL.SDL_GetWindowFlags(game.Window.Handle) & SDL.SDL_WindowFlags.SDL_WINDOW_FULLSCREEN) == 0
```

kg: Note that from poking at it with sharplab, modern .NET optimizes this HasFlag call and doesn't allocate because the allocation isn't observable. But I can't easily prove that NET4x has the same optimization. If you look at the generated IL for a HasFlag call it should have `box` opcodes in it, i.e.:
```msil
IL_0007: box C/A
IL_000c: call instance bool [System.Runtime]System.Enum::HasFlag(class [System.Runtime]System.Enum)
```

src/FNAPlatform/SDL3_FNAPlatform.cs
7aGiven: Thank you, I add comment now.

src/FNAPlatform/SDL3_FNAPlatform.cs
7aGiven: Thank you, I have not known this knowledge. I change code now.

7aGiven:
I test code on VirtualBox MacOS Catalina. It not draw while live resize. So there is no this issue for MacOS.

But I'm not sure is it a bug that no draw while live resize.

src/FNAPlatform/SDL3_FNAPlatform.cs
7aGiven: https://learn.microsoft.com/en-us/visualstudio/profiling/performance-insights-enum-hasflag?view=visualstudio

src/FNAPlatform/SDL3_FNAPlatform.cs
7aGiven: Accoring to https://github.com/libsdl-org/SDL/blob/release-3.4.10/src/video/SDL_video.c#L3035

I not use SDL_WINDOWPOS_CENTERED_DISPLAY. I commit code to calc position by myself now.

PR #591, opened 2026-06-15
Speed up Keyboard.cs

7aGiven:
List\<Keys\>.Contain may slow. So speed up it.

### DIFF
--- a/src/FNAPlatform/SDL2_FNAPlatform.cs
+++ b/src/FNAPlatform/SDL2_FNAPlatform.cs
@@ -976,16 +976,16 @@ namespace Microsoft.Xna.Framework
 				if (evt.type == SDL.SDL_EventType.SDL_KEYDOWN)
 				{
 					Keys key = ToXNAKey(ref evt.key.keysym);
-					if (!Keyboard.keys.Contains(key))
+					if (Keyboard.keys.IsKeyUp(key))
 					{
-						Keyboard.keys.Add(key);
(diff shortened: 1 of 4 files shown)


flibitijibibo:
CC'ing @madewokherd to make sure this doesn't break Nuclex in some way

7aGiven:
You are right. The PR call AddPressedKey and RemovePressedKey when message loop. It may bad for hook app.

madewokherd:
Unfortunately, I do not remember what the name of the game was that used the Nuclex library. AFAICT this doesn't functionally change anything, and the only thing I could imagine being a problem is changing method visibility from private to internal. But the query is probably just "non-public", and I'd expect XNA probably has these as internal as well.

madewokherd:
Found it, it was Revolver360 Re:Actor. I'll test it and get back to you.

7aGiven:
> Unfortunately, I do not remember what the name of the game was that used the Nuclex library. AFAICT this doesn't functionally change anything, and the only thing I could imagine being a problem is changing method visibility from private to internal. But the query is probably just "non-public", and I'd expect XNA probably has these as internal as well.

I checked now. Method visibility is internal for AddPressedKey and RemovePressedKey in XNA.

flibitijibibo:
I'll trust that it was checked with System.Reflection.BindingFlags and not some _other_ way... will do a final review when I'm back in the office.

PR #592, opened 2026-06-16
Add PollEvent between Sleep and SpinWait

7aGiven:
Use SDL_WaitEventTimeout instead of Thread.Sleep

Maybe used for #469

### DIFF
--- a/src/Game.cs
+++ b/src/Game.cs
@@ -447,6 +447,8 @@ namespace Microsoft.Xna.Framework
 
 				/* Now that we have slept into the sleep precision threshold, we need to wait
 				 * for just a little bit longer until the target elapsed time has been reached.
+				 * SpinWait(1) works by pausing the thread for very short intervals, so it is
+				 * an efficient and time-accurate way to wait out the rest of the time.
 				 */
 				while (accumulatedElapsedTime < TargetElapsedTime)
 				{
--- a/src/Game.cs
+++ b/src/Game.cs
@@ -432,6 +432,7 @@ namespace Microsoft.Xna.Framework
 				 */
 				while (accumulatedElapsedTime + worstCaseSleepPrecision < TargetElapsedTime)
 				{
+					System.Threading.Thread.SpinWait(1);
 					System.Threading.Thread.Sleep(1);
 					TimeSpan timeAdvancedSinceSleeping = AdvanceElapsedTime();
 					UpdateEstimatedSleepPrecision(timeAdvancedSinceSleeping);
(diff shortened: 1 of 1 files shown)


7aGiven:
XNA not sleep in main loop.

XNA do message loop all time.

CPU0 is 100% utilization for Test case in XNA on my computer.

Test Case:
```C#
            using (Game g = new Game())
                g.Run();
```

flibitijibibo:
We probably should expose a non-sleeping path for absolute accuracy, but for energy efficiency reasons it'd be best to avoid it by default.

I also worry that WaitEventTimeout doesn't have the same level of precision as our spinwait hybrid system, so that will need to be investigated as well.

7aGiven:
For WaitEventTimeout information:

MacOS: https://developer.apple.com/documentation/appkit/nsapplication/nextevent(matching:until:inmode:dequeue:)?language=objc

Windows: https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-msgwaitformultipleobjects

Linux: ppoll or poll

7aGiven:
> We probably should expose a non-sleeping path for absolute accuracy, but for energy efficiency reasons it'd be best to avoid it by default.

I agree, it is the same to XNA behavior.

7aGiven:
On Windows. Timer resolution is SDL_HINT_TIMER_RESOLUTION. default 1ms for SDL.

Link from SDL docs: http://randomascii.wordpress.com/2013/07/08/windows-timer-resolution-megawatts-wasted/

All win32 timeout api (include sleep) depend on system clock timer resolution. So the PR no accuracy issue on Windows.

7aGiven:
I think there is no accuracy issue for Windows, MacOS, Linux. It is not accuracy for SDL_PLATFORM_XBOXONE and SDL_PLATFORM_XBOXSERIES.

But SDL_WaitEventTimeout seem only implement for Windows, X11, Wayland, cocoa.

Which platforms FNA should consider?

7aGiven:
This PR may not a good idea for Console.

7aGiven:
> We probably should expose a non-sleeping path for absolute accuracy, but for energy efficiency reasons it'd be best to avoid it by default.
> 
> I also worry that WaitEventTimeout doesn't have the same level of precision as our spinwait hybrid system, so that will need to be investigated as well.

@flibitijibibo I agree. The precision of WaitEventTimeout is not good. So I change SpinWait to PollEvent loop. How do you think?

flibitijibibo:
This makes sense - I think I just need to check and make sure that PumpEvents is still happening during those extra Polls, otherwise we're doing extra churn by accident here.

7aGiven:
> This makes sense - I think I just need to check and make sure that PumpEvents is still happening during those extra Polls, otherwise we're doing extra churn by accident here.

Yes, PumpEvent is called when event queue is empty: https://github.com/libsdl-org/SDL/blob/release-3.4.12/src/events/SDL_events.c#L1696

PollEvent just call SDL_WaitEventTimeoutNS(event, 0): https://github.com/libsdl-org/SDL/blob/release-3.4.12/src/events/SDL_events.c#L1552

flibitijibibo:
The SDL_WaitEventTimeoutNS does worry me, doesn't that ultimately replace the spinwait with a CPU yield? That'd be the opposite of what we want in this case.

7aGiven:
> The SDL_WaitEventTimeoutNS does worry me, doesn't that ultimately replace the spinwait with a CPU yield? That'd be the opposite of what we want in this case.

I test SpinWait. But it seem the same to `while (true);`. I just know that _mm_pause but I don't know why it no effect.

7aGiven:
> The SDL_WaitEventTimeoutNS does worry me, doesn't that ultimately replace the spinwait with a CPU yield? That'd be the opposite of what we want in this case.

I agree. I revert the change. Add PollEvent between Sleep and SpinWait.

PR #596, opened 2026-06-30
keyMap KP_PERIOD map to Keys.Decimal and add Keys.VolumeMute support

7aGiven:
SDL_SCANCODE_MUTE SDLK_KP_PERIOD SDLK_KP_DECIMAL


flibitijibibo:
Changes look okay, but the commit message could be more descriptive to explain why these changes are being made.

7aGiven:
KP_PERIOD should map to Keys.Decimal

SDLK_KP_DECIMAL isn't used in any Operate System

I add Keys.VolumeMute. The key exist on my keyboard.

flibitijibibo (approved):
Will merge after tagging 26.07, to make sure we have time to check against as many keyboard layouts as possible

flibitijibibo:
Double checked against XNA and confirmed that Decimal is the correct Keys value.

PR #598, opened 2026-06-30
revert #593 that bug while CapsLock. Ctrl+V without LeftAlt

7aGiven:
KeyMod include CapsLock, ScrollLock, NumLock. Ctrl+V not trigger while they locking.

I test that AltGr mean Keys.LeftCtrl and Keys.LeftAlt in XNA. So when test LeftAlt whether down.


7aGiven:
link #593

PR #604, opened 2026-07-18
optimize: Duplicate DrawString

7aGiven:

Try to unify string class and StringBuilder class.

Core code:
```C#
		internal interface IReadOnlyCharList
		{
			int Length { get; }
			char this[int index] { get; }
		}
		internal struct StringView : IReadOnlyCharList
		{
			private readonly string view;
			internal StringView(string view)
			{
				this.view = view;
			}
			public int Length
			{
				get
				{
					return view.Length;
				}
			}
			public char this[int index]
			{
				get
				{
					return view[index];
				}
			}
		}
		internal struct StringBuilderView : IReadOnlyCharList
		{
			private readonly StringBuilder view;
			internal StringBuilderView(StringBuilder view)
			{
				this.view = view;
			}
			public int Length
			{
				get
				{
					return view.Length;
				}
			}
			public char this[int index]
			{
				get
				{
					return view[index];
				}
			}
		}
```

### DIFF
--- a/src/Graphics/SpriteBatch.cs
+++ b/src/Graphics/SpriteBatch.cs
@@ -712,7 +712,7 @@ namespace Microsoft.Xna.Framework.Graphics
 			{
 				throw new ArgumentNullException("text");
 			}
-			SpriteFont.IReadOnlyCharList view = new SpriteFont.StringBuilderView(text);
+			SpriteFont.StringBuilderView view = new SpriteFont.StringBuilderView(text);
 			DrawString(spriteFont, ref view, position, color, rotation, origin, scale, effects, layerDepth);
 		}
 
@@ -774,7 +774,7 @@ namespace Microsoft.Xna.Framework.Graphics
 			{
 				throw new ArgumentNullException("text");
 			}
-			SpriteFont.IReadOnlyCharList view = new SpriteFont.StringView(text);
+			SpriteFont.StringView view = new SpriteFont.StringView(text);
 			DrawString(spriteFont, ref view, position, color, rotation, origin, scale, effects, layerDepth);
 		}
 
@@ -782,9 +782,9 @@ namespace Microsoft.Xna.Framework.Graphics
 
 		#region Private Methods
 
-		private void DrawString(
+		private void DrawString<T>(
 			SpriteFont spriteFont,
-			ref SpriteFont.IReadOnlyCharList text,
+			ref T text,
 			Vector2 position,
 			Color color,
 			float rotation,
@@ -792,7 +792,7 @@ namespace Microsoft.Xna.Framework.Graphics
 			Vector2 scale,
 			SpriteEffects effects,
 			float layerDepth
-		)
+		) where T : SpriteFont.IReadOnlyCharList
 		{
 			CheckBegin("DrawString");
 			if (text.Length == 0)
--- a/src/Graphics/SpriteBatch.cs
+++ b/src/Graphics/SpriteBatch.cs
@@ -708,16 +708,93 @@ namespace Microsoft.Xna.Framework.Graphics
 			SpriteEffects effects,
 			float layerDepth
 		) {
-			/* FIXME: This method is a duplicate of DrawString(string)!
-			 * The only difference is how we iterate through the StringBuilder.
-			 * We don't use ToString() since it generates garbage.
-			 * -flibit
-			 */
-			CheckBegin("DrawString");
 			if (text == null)
 			{
 				throw new ArgumentNullException("text");
 			}
+			SpriteFont.IReadOnlyCharList view = new SpriteFont.StringBuilderView(text);
+			DrawString(spriteFont, ref view, position, color, rotation, origin, scale, effects, layerDepth);
+		}
+
+		public void DrawString(
+			SpriteFont spriteFont,
+			string text,
+			Vector2 position,
+			Color color
+		) {
+			DrawString(
+				spriteFont,
+				text,
+				position,
+				color,
+				0.0f,
+				Vector2.Zero,
+				Vector2.One,
+				SpriteEffects.None,
+				0.0f
+			);
+		}
+
+		public void DrawString(
+			SpriteFont spriteFont,
+			string text,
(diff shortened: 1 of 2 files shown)


src/Graphics/SpriteFont.cs
```diff
@@ -122,18 +122,9 @@ internal SpriteFont(
...
-			/* FIXME: This method is a duplicate of MeasureString(StringBuilder)!
-			 * The only difference is how we iterate through the string.
-			 * -flibit
-			 */
-			if (text == null)
-			{
-				throw new ArgumentNullException("text");
-			}
+		#region Internal MeasureString Method
+		internal Vector2 MeasureString(ref IReadOnlyCharList text) {
```
flibitijibibo: nitpick:


```suggestion
		#region Internal MeasureString Method

		internal Vector2 MeasureString(ref IReadOnlyCharList text)
		{
```

src/Graphics/SpriteFont.cs
```diff
@@ -216,103 +209,82 @@ public Vector2 MeasureString(string text)
 
 			return result;
 		}
+		#endregion
```
flibitijibibo:
```suggestion

		#endregion
```

flibitijibibo (commented):
Mostly focused on style for this pass but this seems okay; since the reader is a struct I don't anticipate additional memory pressure, but I'll ask others to see if this has any measurable effect on performance.

src/Graphics/SpriteFont.cs
```diff
@@ -217,102 +212,82 @@ public Vector2 MeasureString(string text)
...
 
-			for (int i = 0; i < text.Length; i += 1)
+		#region Internal IReadOnlyCharList Structs
+		internal interface IReadOnlyCharList
+		{
+			int Length { get; }
+			char this[int index] { get; }
+		}
+		internal struct StringView : IReadOnlyCharList
```
kg, line 246: Making this a struct is pointless since you're going to box it anyway, it would be better as a `sealed class` to read how it's going to behave

src/Graphics/SpriteFont.cs
```diff
@@ -122,18 +122,11 @@ internal SpriteFont(
 
 		#endregion
 
-		#region Public MeasureString Methods
+		#region Internal MeasureString Method
 
-		public Vector2 MeasureString(string text)
+		internal Vector2 MeasureString(ref IReadOnlyCharList text)
```
kg: `ref` just adds overhead here because the argument is of a boxed type (an interface). If you had `MeasureString<T>(ref T text) where T : IReadOnlyCharList` then `ref` would make sense and you could avoid boxing.

src/Graphics/SpriteFont.cs
```diff
@@ -122,18 +122,11 @@ internal SpriteFont(
...
 		{
-			/* FIXME: This method is a duplicate of MeasureString(StringBuilder)!
-			 * The only difference is how we iterate through the string.
-			 * -flibit
-			 */
-			if (text == null)
-			{
-				throw new ArgumentNullException("text");
-			}
+			// FIXME: change argument to `(in IReadOnlyCharList text)`. that only allow above C# 7.2
```
kg: `in` is actually worse than `ref` for performance, it tends to generate defensive copies

src/Graphics/SpriteBatch.cs
```diff
@@ -708,149 +708,12 @@ public void DrawString(
...
-					(int) effects
-				);
-
-				/* Add the character width and right-side
-				 * bearing to the line width.
-				 */
-				curOffset.X += cKern.Y + cKern.Z;
-			}
+			SpriteFont.IReadOnlyCharList view = new SpriteFont.StringBuilderView(text);
+			DrawString(spriteFont, ref view, position, color, rotation, origin, scale, effects, layerDepth);
```
kg: As mentioned in another comment, by construction this boxes the `StringBuilderView` to store it in a reference-typed (`IReadOnlyCharList`) local, then passes a reference to the reference, which is unnecessary. Either make `DrawString` generic so you can pass structs into it by-ref or drop the `ref` here.

kg:
Thanks for doing this refactoring. Have you tested it using VS's memory profiler to see whether it allocates? It's possible the modern .NET JIT is smart enough to optimize out some of the boxing here but I would expect it to box each adapter instance.

src/Graphics/SpriteFont.cs
7aGiven: Thank you, I have not get the knowledge before. I know it now from https://giannisakritidis.com/blog/Avoid-Struct-Boxing/

src/Graphics/SpriteFont.cs
7aGiven: Thank you. I follow your suggestion to removed the comment now.

src/Graphics/SpriteFont.cs
7aGiven: Thank you, I follow your suggestion that use generic method instead of interface argument. So no boxing now.

src/Graphics/SpriteBatch.cs
7aGiven: Thank you, I follow your suggestion that use generic method instead of interface argument. So no boxing now.

7aGiven:
> Thanks for doing this refactoring. Have you tested it using VS's memory profiler to see whether it allocates? It's possible the modern .NET JIT is smart enough to optimize out some of the boxing here but I would expect it to box each adapter instance.

I find that there is box instruction in dotnet IL. So I think it must be fix regardless of whether the JIT can optimize it.

kg:
I can't easily test this locally because I don't have any spritefont-based games lying around but the new code looks good.

flibitijibibo (approved):
Overall things look good - it'd be good to get a performance test on this though; [Rogue Legacy](https://github.com/flibitijibibo/RogueLegacy1)'s credits screen is a good test but we can also just test drawing really really really long strings to see how much CPU time it takes before/after.

7aGiven:
Resolve Conflicts

PR #607, opened 2026-07-19
fix LaunchParameters

7aGiven:

### DIFF
--- a/src/LaunchParameters.cs
+++ b/src/LaunchParameters.cs
@@ -17,28 +17,6 @@ namespace Microsoft.Xna.Framework
 {
 	public class LaunchParameters : Dictionary<string, string>
 	{
-		/* FIXME: This whole parser is one big assumption!
-		 *
-		 * I basically started with what MS programs usually accept as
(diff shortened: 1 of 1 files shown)


flibitijibibo:
This commit needs an explanation, test case, _something_ that shows what this does differently and fixes.

7aGiven:
### The PR change the behavior to the same as XNA.
Test case:
```C#
using System;
using System.Collections.Generic;

namespace ConsoleApp4
{
    internal class testfile
    {
        [STAThread]
        {
            foreach (KeyValuePair<string, string> pair in new LaunchParameters())
                Console.Error.WriteLine("key=" + pair.Key.Length + "|" + pair.Key + ", value=" + pair.Value.Length + "|" + pair.Value + "|");
        }
    }
}
```

Arguments with `/arg1:b2 /a /arg2: /:arg3`
XNA output:
```
key=4|arg1, value=2|b2|
key=1|a, value=0||
key=4|arg2, value=0||
```
FNA output:
```
key=4|arg1, value=2|b2|
```

PR #611, opened 2026-07-21
Fix Design Converter

7aGiven:

### DIFF
--- a/src/Design/ColorConverter.cs
+++ b/src/Design/ColorConverter.cs
@@ -94,6 +94,10 @@ namespace Microsoft.Xna.Framework.Design
 			ITypeDescriptorContext context,
 			IDictionary propertyValues
 		) {
+			if (propertyValues == null)
+			{
+				throw new ArgumentNullException("propertyValues");
+			}
 			return (object) new Color(
 				(int) propertyValues["R"],
 				(int) propertyValues["G"],
--- a/src/Design/ColorConverter.cs
+++ b/src/Design/ColorConverter.cs
@@ -23,7 +23,13 @@ namespace Microsoft.Xna.Framework.Design
 
 		public ColorConverter() : base()
 		{
-			// FIXME: Initialize propertyDescriptions... how? -flibit
+			Type Color = typeof(Color);
+			propertyDescriptions = new PropertyDescriptorCollection(new PropertyDescriptor[] {
+				new PropertyPropertyDescriptor(Color.GetProperty("R")),
+				new PropertyPropertyDescriptor(Color.GetProperty("G")),
+				new PropertyPropertyDescriptor(Color.GetProperty("B")),
+				new PropertyPropertyDescriptor(Color.GetProperty("A"))
+			});
 		}
 
 		#endregion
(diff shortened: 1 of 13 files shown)


7aGiven:
I create a new file, but seem not be commited.

7aGiven:
OK, I append the content to MathTypeConverter.cs

flibitijibibo:
The new types have to be added to the csproj files and Makefile in addition to `git add`ing the new files.

Those changes make me wonder if there's an easier way to do that though, because I would have expected there to be built-in classes to do that for Design namespaces in general... this might be a question for @kg.

7aGiven:
> Those changes make me wonder if there's an easier way to do that though, because I would have expected there to be built-in classes to do that for Design namespaces in general... this might be a question for @kg.

Which range of code do you want to simply?

7aGiven:
I commit new code that allow `new ColorConverter().ConvertFromString("0x1D, 0X3E, &H4F, &h5C")`.

It's the same behavior as XNA.

Optimize ConvertTo string and ConvertFrom string.

7aGiven:
Add a commit.
```C#
Rectangle rectangle = new Rectangle { X = 1, Y = 2, Width = 3, Height = 4 };
Console.WriteLine(new ColorConverter().ConvertToString(rectangle));
```
XNA output: `{X:1 Y:2 Width:3 Height:4}`
FNA throw System.InvalidCastException

It's the same as typeof(InstanceDescriptor)

src/Design/ColorConverter.cs
```diff
@@ -38,15 +44,8 @@ object value
...
-				string[] v = s.Split(
-					culture.TextInfo.ListSeparator.ToCharArray()
-				);
-				return new Color(
-					int.Parse(v[0], culture),
-					int.Parse(v[1], culture),
-					int.Parse(v[2], culture),
-					int.Parse(v[3], culture)
-				);
+				StringListEnumerator<int> enumerator = new StringListEnumerator<int>(culture, s);
```
kg, line 47: What is StringListEnumerator?

kg: Nevermind, I see it's defined later.

src/Design/MathTypeConverter.cs
```diff
@@ -39,22 +34,14 @@ public override bool CanConvertFrom(
...
 		public override bool CanConvertTo(
 			ITypeDescriptorContext context,
 			Type destinationType
 		) {
-			if (supportStringConvert && destinationType == typeof(string))
-			{
-				return true;
-			}
-			return base.CanConvertTo(context, destinationType);
+			return destinationType == typeof(InstanceDescriptor) || base.CanConvertTo(context, destinationType);
```
kg, line 44: This was originally `== typeof(string)` and now it's `== typeof(InstanceDescriptor)`. I'm guessing the original code was a copy-paste error?

7aGiven: TypeConverter have implemented ConvertTo string. So base.CanConvertTo return true when destinationType == typeof(string). Sure, TypeConverter just object.ToString in ConvertTo for typeof(string).

reference: https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.typeconverter.canconvertto?view=netframework-4.8.1

7aGiven:
new commit that when new CultureInfo("de-DE"), ConvertTo typeof(string), float field should 1.123f to "1,123"

kg (approved):

kg:
I don't know much about the design namespace but the code looks good

PR #615, opened 2026-07-21
Fix MathHelper

7aGiven:

Optimize Clamp.

fix Max and Min float.NaN issue.

Make the same accuracy as XNA for ToDegrees and ToRadians.

### DIFF
--- a/src/MathHelper.cs
+++ b/src/MathHelper.cs
@@ -316,7 +316,6 @@ namespace Microsoft.Xna.Framework
 		/// <param name="degrees">The angle in degrees.</param>
 		/// <returns>The angle in radians.</returns>
 		/// <remarks>
-		/// This method uses double precision internally, though it returns single float.
 		/// Factor = pi / 180
 		/// </remarks>
 		public static float ToRadians(float degrees)
--- a/src/MathHelper.cs
+++ b/src/MathHelper.cs
@@ -149,11 +149,14 @@ namespace Microsoft.Xna.Framework
 		/// <returns>The clamped value.</returns>
 		public static float Clamp(float value, float min, float max)
 		{
-			// First we check to see if we're greater than the max.
-			value = (value > max) ? max : value;
-
-			// Then we check to see if we're less than the min.
-			value = (value < min) ? min : value;
+			if (value > max) // First we check to see if we're greater than the max.
+			{
+				value = max;
+			}
+			else if (value < min) // Then we check to see if we're less than the min.
+			{
+				value = min;
+			}
 
 			// There's no check to see if min > max.
 			return value;
@@ -245,7 +248,15 @@ namespace Microsoft.Xna.Framework
 		/// <returns>The greater value.</returns>
 		public static float Max(float value1, float value2)
 		{
-			return value1 > value2 ? value1 : value2;
+			if (value1 > value2)
+			{
+				return value1;
+			}
+			if (float.IsNaN(value1))
+			{
+				return value1;
+			}
+			return value2;
 		}
 
 		/// <summary>
@@ -256,7 +267,15 @@ namespace Microsoft.Xna.Framework
 		/// <returns>The lesser value.</returns>
 		public static float Min(float value1, float value2)
 		{
-			return value1 < value2 ? value1 : value2;
+			if (value1 < value2)
+			{
+				return value1;
+			}
+			if (float.IsNaN(value1))
+			{
+				return value1;
+			}
+			return value2;
 		}
 
 		/// <summary>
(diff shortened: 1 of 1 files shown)


flibitijibibo:
The commits being pushed need to be improved - I've been covering a lot of them but let's make them thorough from the start:

- The first commit should only fix the IsNan cases with a test case in the message showing different results before/after
- The second commit should only change the precision of ToDegrees/ToRadians, with a test case in the message showing different results before/after

7aGiven:
Test Case:
```C#
using System;

namespace ConsoleApp5
{
    internal class Show
    {
        [STAThread]
        {
            Console.WriteLine("Min:" + MathHelper.Min(float.NaN, 1) + " | " + MathHelper.Min(1, float.NaN));
            Console.WriteLine("Max:" + MathHelper.Max(float.NaN, 1) + " | " + MathHelper.Max(1, float.NaN));
            Console.WriteLine("ToDegrees:" + MathHelper.ToDegrees(1.3f) + " | " + MathHelper.ToDegrees(1.5f));
            Console.WriteLine("ToRadians:" + MathHelper.ToRadians(1.69f) + " | " + MathHelper.ToRadians(1.769f));
        }
    }
}

```
XNA output:
```
Min:NaN | NaN
Max:NaN | NaN
ToDegrees:74.4845 | 85.94366
ToRadians:0.02949606 | 0.03087487
```
FNA output:
```
Min:1 | NaN
Max:1 | NaN
ToDegrees:74.48451 | 85.94367
ToRadians:0.02949607 | 0.03087488
```

src/MathHelper.cs
```diff
@@ -284,12 +303,11 @@ public static float SmoothStep(float value1, float value2, float amount)
 		/// <param name="radians">The angle in radians.</param>
 		/// <returns>The angle in degrees.</returns>
 		/// <remarks>
-		/// This method uses double precision internally, though it returns single float.
 		/// Factor = 180 / pi
 		/// </remarks>
 		public static float ToDegrees(float radians)
 		{
-			return (float) (radians * 57.295779513082320876798154814105);
+			return radians * 57.2957764f;
```
flibitijibibo: Double checked this factor constant, I'm getting 57.295779513?

7aGiven: Although it's strange, but the const is from XNA dll.

flibitijibibo: We'll need a clean-room method of getting this number... I wonder if there's some Pi constant we can derive this from?

7aGiven: I know it now. It is a float divide Precision issue. I write the new code for the same result.

7aGiven:
New commit with clean-room method.

flibitijibibo:
New version looks much better! In the meantime, someone brought this up in Discord for the other parts:

> Testing explicitly for NaN seems unnecessary, one could just invert the condition

7aGiven:
> Testing explicitly for NaN seems unnecessary, one could just invert the condition

I can't understand, What it mean?

src/MathHelper.cs
```diff
@@ -256,7 +267,15 @@ public static float Max(float value1, float value2)
...
-			return value1 < value2 ? value1 : value2;
+			if (value1 < value2)
+			{
+				return value1;
+			}
+			if (float.IsNaN(value1))
+			{
+				return value1;
+			}
+			return value2;
```
flibitijibibo, line 278: If I'm understanding them, it means we can just do this instead:


```suggestion
			return value1 > value2 ? value2 : value1;
```

By switching just the comparison operator it handles the NaN case in the direction that we want.

7aGiven: I test the change:
```C#
		/// <summary>
		/// Returns the greater of two values.
		/// </summary>
		/// <param name="value1">Source value.</param>
		/// <param name="value2">Source value.</param>
		/// <returns>The greater value.</returns>
		public static float Max(float value1, float value2)
		{
			return value1 < value2 ? value2 : value1;
		}

		/// <summary>
		/// Returns the lesser of two values.
		/// </summary>
		/// <param name="value1">Source value.</param>
		/// <param name="value2">Source value.</param>
		/// <returns>The lesser value.</returns>
		public static float Min(float value1, float value2)
		{
			return value1 > value2 ? value2 : value1;
		}
```
The test case output:
```
Min:NaN | 1
Max:NaN | 1
```
It just swap NaN and 1.

Addition info: for any float operator, if exist NaN argument, result is false.

flibitijibibo: Checked with Discord again and confirmed this, so I _think_ the current PR is right. Will do one final runthrough tomorrow morning.

flibitijibibo: Scratch that, IEEE says to do this, so we're good

PR #617, opened 2026-07-23
Fix FNAWindow.Title is null

7aGiven:


flibitijibibo (changes requested):
Since we control the internal constructor interface, let's just add the title as a parameter.

PR #619, opened 2026-07-23
Remove Game.Components element when GameComponent Disposing. Fix OnUpdateOrderChanged and OnEnabledChanged when inherit GameComponent.

7aGiven:

### DIFF
--- a/src/GameComponent.cs
+++ b/src/GameComponent.cs
@@ -34,10 +34,6 @@ namespace Microsoft.Xna.Framework
 				if (_enabled != value)
 				{
 					_enabled = value;
-					if (this.EnabledChanged != null)
-					{
-						this.EnabledChanged(this, EventArgs.Empty);
-					}
 					OnEnabledChanged(this, null);
 				}
 			}
@@ -54,10 +50,6 @@ namespace Microsoft.Xna.Framework
 				if (_updateOrder != value)
 				{
 					_updateOrder = value;
-					if (this.UpdateOrderChanged != null)
-					{
-						this.UpdateOrderChanged(this, EventArgs.Empty);
-					}
 					OnUpdateOrderChanged(this, null);
 				}
 			}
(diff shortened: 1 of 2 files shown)


src/GameComponent.cs
```diff
@@ -121,18 +113,35 @@ public virtual void Update(GameTime gameTime) {}
 
 		#region Protected Virtual Methods
 
-		protected virtual void OnUpdateOrderChanged(object sender, EventArgs args) {}
+		protected virtual void OnUpdateOrderChanged(object sender, EventArgs args) {
```
flibitijibibo:
```suggestion
		protected virtual void OnUpdateOrderChanged(object sender, EventArgs args)
		{
```

src/GameComponent.cs
```diff
@@ -121,18 +113,35 @@ public virtual void Update(GameTime gameTime) {}
...
+			if (this.UpdateOrderChanged != null)
+			{
+				this.UpdateOrderChanged(this, EventArgs.Empty);
+			}
+		}
 
-		protected virtual void OnEnabledChanged(object sender, EventArgs args) {}
+		protected virtual void OnEnabledChanged(object sender, EventArgs args) {
```
flibitijibibo:
```suggestion
		protected virtual void OnEnabledChanged(object sender, EventArgs args)
		{
```

src/GameComponent.cs
```diff
@@ -121,18 +113,35 @@ public virtual void Update(GameTime gameTime) {}
...
 		protected virtual void Dispose(bool disposing)
 		{
-			if (disposing && Disposed != null)
+			if (disposing)
 			{
-				Disposed(this, EventArgs.Empty);
+				if (Game != null)
+				{
+					Game.Components.Remove(this);
+				}
```
flibitijibibo, line 142: Does this not happen elsewhere? This is where a commit message specifically for this change would help understand the full picture of why this is important.

flibitijibibo (commented):
Callback changes look good, just a couple minor style things and a missing message for the second change.

src/GameComponent.cs
7aGiven: What do you mean "Does this not happen elsewhere?" ?

Commit message? Do you mean "Remove Game.Components element when GameComponent Disposing." in title?

flibitijibibo: Took a look at GameComponentCollection just to be sure, I think we can test this by making a Game with a ComponentRemoved callback hooked, then create and Dispose a component. If this is right, XNA should trigger ComponentRemoved while FNA's current revision won't.

7aGiven:
Test case:
```C#
using System;

namespace ConsoleApp3
{
    internal static class Program
    {
        [STAThread]
        private static unsafe void Main(string[] args)
        {
            using (Game g = new FNAGame())
                g.Run();
        }
    }
    class FNAGame : Game
    {
        private readonly GameComponent component;
        internal FNAGame()
        {
            new GraphicsDeviceManager(this);
            Components.ComponentAdded += Components_ComponentAdded;
            Components.ComponentRemoved += Components_ComponentRemoved;
            component = new GameComponent(this);
            Components.Add(component);

        }

        private void Components_ComponentAdded(object sender, GameComponentCollectionEventArgs e)
        {
            Console.WriteLine("Components_ComponentAdded");
        }

        private void Components_ComponentRemoved(object sender, GameComponentCollectionEventArgs e)
        {
            Console.WriteLine("Components_ComponentRemoved");
        }

        bool b = true;
        protected override void Update(GameTime gameTime)
        {
            base.Update(gameTime);
            if (b)
            {
                b = false;
                component.Dispose();
            }
        }
    }
}
```
XNA output:
```
Components_ComponentAdded
Components_ComponentRemoved
```
FNA output:
```
Components_ComponentAdded
```

PR #621, opened 2026-07-23
Optimize Equals and fix GetHashCode for CurveKey

7aGiven:

### DIFF
--- a/src/CurveKey.cs
+++ b/src/CurveKey.cs
@@ -160,7 +160,12 @@ namespace Microsoft.Xna.Framework
 
 		public bool Equals(CurveKey other)
 		{
-			return (this == other);
+			return other != null &&
+				other.Position == Position &&
+				other.Value == Value &&
+				other.TangentIn == TangentIn &&
+				other.TangentOut == TangentOut &&
+				other.Continuity == Continuity;
 		}
 
 		#endregion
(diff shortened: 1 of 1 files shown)


7aGiven:
When call Equals, this must not null.
When call operator ==, maybe a == null b == null.

So move code from operator == to Equals.

7aGiven:
The change for GetHashCode is for the same behavior as XNA.

flibitijibibo (commented):
Equals changes look good. I'm surprised the hashcode is just additions; is there a test case where we can show matching hashcodes for two different CurveKeys where we just swap the numbers for TangentIn and TangentOut?

7aGiven:
```C#
using System;

namespace ConsoleApp5
{
    internal class Show
    {
        [STAThread]
        {
            Console.WriteLine(new CurveKey(1.1f, 1.2f, 1.3f, 1.4f, CurveContinuity.Step).GetHashCode().ToString("X"));
        }
    }
}

```
XNA output: `FE800001`
FNA output: `3`

7aGiven:
> Equals changes look good. I'm surprised the hashcode is just additions; is there a test case where we can show matching hashcodes for two different CurveKeys where we just swap the numbers for TangentIn and TangentOut?

I can't understand. swap argument for add and xor is the same result. So why test it?

flibitijibibo:
> > Equals changes look good. I'm surprised the hashcode is just additions; is there a test case where we can show matching hashcodes for two different CurveKeys where we just swap the numbers for TangentIn and TangentOut?
> 
> I can't understand. swap argument for add and xor is the same result. So why test it?

That specific example might not make a difference, was mostly trying to think of _something_ that could show why this hashcode uses + instead of ^ like a lot of the others do. We don't have great tests for _any_ of  the hashcode calculators so we usually only determine it based on a specific title that made use of them in some way.

7aGiven:
> That specific example might not make a difference, was mostly trying to think of _something_ that could show why this hashcode uses + instead of ^ like a lot of the others do. We don't have great tests for _any_ of the hashcode calculators so we usually only determine it based on a specific title that made use of them in some way.

For the same Integer value, ^ result is 0, + result is add.

flibitijibibo:
Took a look and it turns out that while ^ is used for HashCode implementations, the math types are a notable exception, so this actually makes it more consistent with the others. I'll take your word for it on this one :)

7aGiven:
> Took a look and it turns out that while ^ is used for HashCode implementations, the math types are a notable exception, so this actually makes it more consistent with the others. I'll take your word for it on this one :)

I think the reason is that ^ result is 0 for the same interger. So if field exist the same value, HashCode will be 0. It is bad.

7aGiven:
Addition info: `System.Int.GetHashCode` return self. `System.Single.GetHashCode` return `*(int*)&f`

PR #622, opened 2026-07-23
Fix TouchCollection

7aGiven:

### DIFF
--- a/src/Input/Touch/TouchCollection.cs
+++ b/src/Input/Touch/TouchCollection.cs
@@ -52,12 +52,15 @@ namespace Microsoft.Xna.Framework.Input.Touch
 			{
 				if (touches == null)
 					throw new ArgumentOutOfRangeException();
+				if (index < 0 || index >= this.Count)
+				{
+					throw new ArgumentOutOfRangeException("index");
+				}
 				return touches[index];
 			}
 			set
 			{
-				// This will cause a runtime exception
-				touches[index] = value;
+				throw new NotSupportedException();
 			}
 		}
 
@@ -73,6 +76,14 @@ namespace Microsoft.Xna.Framework.Input.Touch
 
 		public TouchCollection(TouchLocation[] touches)
 		{
+			if (touches == null)
+			{
+				throw new ArgumentNullException("touches");
+			}
+			if (touches.Length > 8)
+			{
+				throw new ArgumentOutOfRangeException("touches");
+			}
 			this.touches = new List<TouchLocation>(touches);
 		}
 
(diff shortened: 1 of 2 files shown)


flibitijibibo (changes requested):
We actually _do_ support touch on various systems (including PC!), so this may need to get re-done. Maybe when no touch devices are present...?

7aGiven:
> We actually _do_ support touch on various systems (including PC!), so this may need to get re-done. Maybe when no touch devices are present...?

I can't understand. My PR mean TouchCollection is readonly and can't change by user. And up to 8-finger multi-touch.

flibitijibibo:
I probably don't understand the Touch namespace enough, will ping the Apple channel about this

src/Input/Touch/TouchCollection.cs
```diff
@@ -73,6 +76,14 @@ public TouchLocation this[int index]
 
 		public TouchCollection(TouchLocation[] touches)
 		{
+			if (touches == null)
+			{
+				throw new ArgumentNullException("touches");
+			}
+			if (touches.Length > 17) // XNA: touches.Length > 8
+			{
+				throw new ArgumentOutOfRangeException("touches");
```
kg, line 83: I don't really understand the value of having this check, it would mean that if you somehow generated 18 touches in production the game would crash. It would be better to drop extra touches if there is some mechanism elsewhere that can't handle 18 touches

7aGiven: Thank you, I apply your suggestion now. And I remove touchs == null check in other methods.

7aGiven: I use Microsoft.Xna.Framework.Input.Touch.TouchPanel.MAX_TOUCHES now.

7aGiven: TouchPanel.GetState have limit 8 multi-touch. It only pass max length = 8 array to TouchCollection.

7aGiven:
I find that IsConnected=true for XNA. SDL_GetTouchDevices(int *count) count =0 in my computer. So FNA report IsConnected=false.

I don't know why.

7aGiven:
Test case:
```C#
using Microsoft.Xna.Framework.Input.Touch;
using System;

namespace ConsoleApp5
{
    internal static class Show
    {
        [STAThread]
        {
            Console.WriteLine(TouchPanel.GetState().IsConnected);
            Console.WriteLine(new TouchCollection().IsConnected);
        }
    }
}
```
XNA output:
```
True
False
```
It mean IsConnected not directly return TouchDeviceExists.

7aGiven:
> I find that IsConnected=true for XNA. SDL_GetTouchDevices(int *count) count =0 in my computer. So FNA report IsConnected=false.
> 
> I don't know why.

My guess is that `SDL_GetTouchDevices(int *count) != NULL` mean `IsConnected = true`.

PR #624, opened 2026-07-24
Fix GetHashCode for VertexElement VertexPositionColor VertexPositionColorTexture VertexPositionNormalTexture VertexPositionTexture GamePadThumbSticks GamePadTriggers

7aGiven:


flibitijibibo:
Are we sure there isn't a better way to get this...? The bitflags we had for the one type seem like a good alternative to getting into GCHandles.

7aGiven:
> Are we sure there isn't a better way to get this...? The bitflags we had for the one type seem like a good alternative to getting into GCHandles.

I apply your suggestion.

PR #625, opened 2026-07-24
Fix Exception for Vector

7aGiven:


flibitijibibo:
The Exceptions are okay but for simplicity's sake let's keep the rest of the code as-is; if there's a measurable performance improvement from the other changes we can do that in a separate PR.

7aGiven:
The rest of code. The caller method call a method that need check argument that caller method not need check them.

So I copy the the method content to caller method.

flibitijibibo:
Got it, that makes more sense!

PR #629, opened 2026-07-25
Fix Exception for AudioCategory AudioEngine Cue SoundBank WaveBank AudioEmitter DynamicSoundEffectInstance Microphone SoundEffect SoundEffectInstance

7aGiven:

### DIFF
--- a/src/Audio/AudioCategory.cs
+++ b/src/Audio/AudioCategory.cs
@@ -78,6 +78,10 @@ namespace Microsoft.Xna.Framework.Audio
 
 		public void SetVolume(float volume)
 		{
+			if (volume < 0f)
+			{
+				throw new ArgumentException("Volume must be a positive float value.");
+			}
 			lock (parent.gcSync)
 			{
 				if (parent.IsDisposed)
--- a/src/Audio/DynamicSoundEffectInstance.cs
+++ b/src/Audio/DynamicSoundEffectInstance.cs
@@ -92,7 +92,7 @@ namespace Microsoft.Xna.Framework.Audio
 			int sampleRate,
 			AudioChannels channels
 		) : base() {
-			if (sampleRate < 8000 || sampleRate > 48000)
+			if (sampleRate < 8000 || sampleRate > 384000) // sampleRate < 8000 || sampleRate > 48000
 			{
 				throw new ArgumentOutOfRangeException("sampleRate");
 			}
--- a/src/Audio/DynamicSoundEffectInstance.cs
+++ b/src/Audio/DynamicSoundEffectInstance.cs
@@ -92,7 +92,7 @@ namespace Microsoft.Xna.Framework.Audio
 			int sampleRate,
 			AudioChannels channels
 		) : base() {
-			if (sampleRate < 8000 || sampleRate > 384000) // XNA: sampleRate < 8000 || sampleRate > 48000
+			if (sampleRate < FAudio.FAUDIO_MIN_SAMPLE_RATE || sampleRate > FAudio.FAUDIO_MAX_SAMPLE_RATE) // XNA: sampleRate < 8000 || sampleRate > 48000
 			{
 				throw new ArgumentOutOfRangeException("sampleRate");
 			}
--- a/src/Audio/DynamicSoundEffectInstance.cs
+++ b/src/Audio/DynamicSoundEffectInstance.cs
@@ -92,7 +92,7 @@ namespace Microsoft.Xna.Framework.Audio
 			int sampleRate,
 			AudioChannels channels
 		) : base() {
-			if (sampleRate < 8000 || sampleRate > 384000) // sampleRate < 8000 || sampleRate > 48000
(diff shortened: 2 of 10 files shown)


src/Audio/AudioCategory.cs
```diff
@@ -78,6 +78,10 @@ public void Resume()
 
 		public void SetVolume(float volume)
 		{
+			if (volume < 0f)
```
flibitijibibo, line 81: Oddly enough this is supported by XAudio2, so even though this is an XNA Exception let's comment this to expose more audio capabilities at no cost.

src/Audio/DynamicSoundEffectInstance.cs
```diff
@@ -76,6 +92,14 @@ public DynamicSoundEffectInstance(
 			int sampleRate,
 			AudioChannels channels
 		) : base() {
+			if (sampleRate < 8000 || sampleRate > 48000)
```
flibitijibibo: Same for this check

src/Audio/SoundEffect.cs
```diff
@@ -61,6 +72,10 @@ out result
 			}
 			set
 			{
+				if (value < 0f || value > 1f)
```
flibitijibibo: Here as well

src/Audio/SoundEffect.cs
```diff
@@ -365,29 +398,47 @@ public static TimeSpan GetSampleDuration(
...
-			int ms = (int) (
-				(sizeInBytes / (int) channels) /
-				(sampleRate / 1000.0f)
-			);
-			return new TimeSpan(0, 0, 0, 0, ms);
+			if (sizeInBytes < 0)
+			{
+				throw new ArgumentException("Buffer size cannot be negative.", "sizeInBytes");
+			}
+			if (sampleRate < 8000 || sampleRate > 48000)
```
flibitijibibo: Ditto

src/Audio/SoundEffect.cs
```diff
@@ -365,29 +398,47 @@ public static TimeSpan GetSampleDuration(
...
-				duration.TotalSeconds *
-				sampleRate *
-				(int) channels *
-				2 // 16-bit PCM!
-			);
+			if (duration.TotalMilliseconds < 0.0 || duration.TotalMilliseconds > int.MaxValue)
+			{
+				throw new ArgumentOutOfRangeException("duration");
+			}
+			if (sampleRate < 8000 || sampleRate > 48000)
```
flibitijibibo: Ditto

src/Audio/SoundEffectInstance.cs
```diff
@@ -129,6 +142,14 @@ public float Volume
 			}
 			set
 			{
+				if (IsDisposed)
+				{
+					throw new ObjectDisposedException(GetType().Name, "This object has already been disposed.");
+				}
+				if (value < 0f || value > 1f)
```
flibitijibibo: Ditto

flibitijibibo (changes requested):
Most of these are good but some we opted out of since it doesn't really protect from anything, and just adds arbitrary reasons to crash.

7aGiven:
How do you think about sampleRate?

I think it should be 8k Hz ~ 384k Hz.

src/Audio/AudioCategory.cs
7aGiven: FACTVOLUME_MIN = 0.0f

flibitijibibo:
> How do you think about sampleRate?
> 
> I think it should be 8k Hz ~ 384k Hz.

Let's do 1000 to 200000, since those are the XAudio2 constants:

https://github.com/FNA-XNA/FAudio/blob/master/include/FAudio.h#L356

src/Audio/AudioCategory.cs
flibitijibibo: Good catch, thought I was still reviewing the SoundEffect side!

src/Audio/SoundEffect.cs
```diff
@@ -533,6 +584,38 @@ public static SoundEffect FromStream(Stream stream)
...
 		#endregion
 
+		#region Internal Static Methods
+
+		internal static TimeSpan INTERNAL_GetSampleDuration(
+			int sizeInBytes,
+			int sampleRate,
+			AudioChannels channels
+		)
+		{
```
flibitijibibo:
```suggestion
		) {
```

src/Audio/SoundEffect.cs
```diff
@@ -533,6 +584,38 @@ public static SoundEffect FromStream(Stream stream)
...
+			);
+			return new TimeSpan(0, 0, 0, 0, ms);
+		}
+
+		internal static int INTERNAL_GetSampleSizeInBytes(
+			TimeSpan duration,
+			int sampleRate,
+			AudioChannels channels
+		)
+		{
```
flibitijibibo:
```suggestion
		) {
```

7aGiven:
pending review

flibitijibibo (approved):
Applying a couple style fixes then merging...

PR #630, opened 2026-07-26
Revert 984feef, Adapter default value

7aGiven:

https://github.com/FNA-XNA/FNA/commit/984feefa2845dd9da21c7b8bcf9ade5d8f617370

### DIFF
--- a/src/GraphicsDeviceInformation.cs
+++ b/src/GraphicsDeviceInformation.cs
@@ -35,7 +35,7 @@ namespace Microsoft.Xna.Framework
 			get { return adapter; }
 			set
 			{
-				if (value == null)
+				if (adapter == null) // The same behavior as XNA
 				{
 					throw new ArgumentNullException("value", "Adapter cannot be null.  Try using GraphicsAdapter.DefaultAdapter instead.");
 				}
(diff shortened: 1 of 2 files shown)


flibitijibibo:
So if you create GraphicsDeviceInformation yourself, setting Adapter to null doesn't break until setting it a second time?

7aGiven:
Test case:
```C#
using System;

namespace ConsoleApp5
{
    internal static class Show
    {
        [STAThread]
        {
            Console.WriteLine(1);
            GraphicsDeviceInformation gdi = new GraphicsDeviceInformation();
            Console.WriteLine(2);
            gdi.Adapter = null;
            Console.WriteLine(3);
            gdi.Adapter = null;
            Console.WriteLine(4);
        }
    }
}
```
XNA output:
```
1
2
3

Unhandled Exception: System.ArgumentNullException: Adapter cannot be null.  Try using GraphicsAdapter.DefaultAdapter instead.
Parameter name: value
   at Microsoft.Xna.Framework.GraphicsDeviceInformation.set_Adapter(GraphicsAdapter value)
   at ConsoleApp5.Show.Main(String[] args) in C:\Users\\source\repos\ConsoleApp2\ConsoleApp2\show.cs:line 16
```

flibitijibibo:
Crazy find, thanks for checking this out!

PR #633, opened 2026-07-26
Fix Exception for AudioEngine SoundBank WaveBank

7aGiven:

### DIFF
--- a/src/Audio/AudioEngine.cs
+++ b/src/Audio/AudioEngine.cs
@@ -11,6 +11,7 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
+using System.IO;
 using System.Runtime.InteropServices;
 #endregion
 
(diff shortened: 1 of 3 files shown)


src/Audio/AudioEngine.cs
```diff
@@ -118,6 +119,14 @@ string rendererId
 			{
 				throw new ArgumentNullException("settingsFile", "This method does not accept null for this parameter.");
 			}
+			using (FileStream stream = File.OpenRead(settingsFile))
```
flibitijibibo: We should be able to grab this from the pointer data below. We may also be able to get it via the HRESULT for loading the engine, but I forget if we implemented that specific error code.

7aGiven: But in below code, read bytes from the whole file.

7aGiven: Addition, the check also need by SoundBank and WaveBank

flibitijibibo: Double checked, we can check for E_INVALIDDATA for all three files https://github.com/FNA-XNA/FAudio/blob/master/include/FACT.h#L49

7aGiven: OK, I can see it return FACTENGINE_E_INVALIDDATA at https://github.com/FNA-XNA/FAudio/blob/26.07/src/FACT_internal.c#L2030

But why it check big endiness and little endiness?

flibitijibibo: We work with Xbox 360 data from time to time, so the data may be partially byteswapped.

PR #634, opened 2026-07-27
Fix recursion in CurveKey Equals

7aGiven:

Fix #621 

call obj == null in operator == cause recursion


flibitijibibo:
Maybe ((object) a) == null?

flibitijibibo:
Forgot about object.ReferenceEquals, yeah that should fix it once CI is happy!

7aGiven:
> Maybe ((object) a) == null?

I think it is confused for other developer

PR #635, opened 2026-07-30
Fix DisplayMode AspectRatio

7aGiven:

### DIFF
--- a/src/Graphics/DisplayMode.cs
+++ b/src/Graphics/DisplayMode.cs
@@ -13,7 +13,6 @@
 
 #region Using Statements
 using System;
-using System.Globalization;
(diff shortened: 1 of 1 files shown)


flibitijibibo:
I get the Height change but what's with the other changes? (Another reminder: Commit messages are important!)

7aGiven:
According to: https://learn.microsoft.com/en-us/previous-versions/windows/xna/bb196197(v=xnagamestudio.40)

DisplayMode not override these methods.

flibitijibibo (changes requested):
Getting these two changes as two separate patches will help with debugging in case someone needs to identify regressions later.

PR #636, opened 2026-07-30
Fix 10a1472

7aGiven:

https://github.com/FNA-XNA/FNA/commit/10a14723f92fb65bae9cb4924b2a5507e97a25a9

The default Name for SoundEffect is string.Empty in XNA.

### DIFF
--- a/src/Audio/SoundEffect.cs
+++ b/src/Audio/SoundEffect.cs
@@ -38,7 +38,6 @@ namespace Microsoft.Xna.Framework.Audio
 			private set;
 		}
 
-		private string name;
 		public string Name
 		{
 			get { return name; }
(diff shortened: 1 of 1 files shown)


src/Audio/SoundEffect.cs
```diff
@@ -38,6 +38,7 @@ public bool IsDisposed
 			private set;
 		}
 
+		private string name;
```
flibitijibibo: This part of the change isn't necessary.

flibitijibibo (changes requested):
Functional change looks good, just need to shrink the total diff size.

7aGiven:
Sometimes private field like INTERNEL_name is above public property. Sometimes it is in region Private Field.

Which should I choose next time?

7aGiven:
The PR fix throw ArgumentNullException("name") due to https://github.com/FNA-XNA/FNA/commit/10a14723f92fb65bae9cb4924b2a5507e97a25a9

7aGiven:
link #629

PR #641, opened 2026-07-31
IsHeadset=true

7aGiven:

IsHeadset is always true in XNA on Windows Phone and Windows.

I'm not sure how Xbox 360 is. How to know it?

### DIFF
--- a/src/Audio/Microphone.cs
+++ b/src/Audio/Microphone.cs
@@ -56,9 +56,9 @@ namespace Microsoft.Xna.Framework.Audio
 			}
 			set
 			{
-				if (	value.TotalMilliseconds < 100.0 ||
-					value.TotalMilliseconds > 1000.0 ||
-					value.TotalMilliseconds % 10.0 != 0	)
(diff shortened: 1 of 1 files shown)


src/Audio/Microphone.cs
```diff
@@ -56,9 +56,9 @@ public TimeSpan BufferDuration
 			}
 			set
 			{
-				if (	value.TotalMilliseconds < 100 ||
-					value.TotalMilliseconds > 1000 ||
-					value.TotalMilliseconds % 10 != 0	)
+				if (	value.TotalMilliseconds < 100.0 ||
+					value.TotalMilliseconds > 1000.0 ||
+					value.TotalMilliseconds % 10.0 != 0	)
```
flibitijibibo:
```suggestion
				if (	value.TotalMilliseconds < 100 ||
					value.TotalMilliseconds > 1000 ||
					value.TotalMilliseconds % 10 != 0	)
```

flibitijibibo:
I think SDL can identify these via device roles but for now we can just match desktop here.

PR #644, opened 2026-07-31
Fix Dispose for SoundBank Cue WaveBank AudioEngine StorageContainer

7aGiven:


7aGiven:
IsDisposed is true when trigger Disposing in XNA.

Disposing is not triggered when GC.

PR #646, opened 2026-07-31
Fix Dispose for DynamicSoundEffectInstance

7aGiven:

>The finalizer implicitly calls [Finalize](https://learn.microsoft.com/en-us/dotnet/api/system.object.finalize) on the base class of the object.

I don't understand why ReRegisterForFinalize. Which instance leak?

### DIFF
--- a/src/Audio/DynamicSoundEffectInstance.cs
+++ b/src/Audio/DynamicSoundEffectInstance.cs
@@ -123,16 +123,6 @@ namespace Microsoft.Xna.Framework.Audio
 
 		#endregion
 
-		#region Destructor
-
-		~DynamicSoundEffectInstance()
-		{
-			// FIXME: ReRegisterForFinalize? -flibit
-			Dispose();
-		}
-
-		#endregion
-
 		#region Public Methods
 
 		public TimeSpan GetSampleDuration(int sizeInBytes)


flibitijibibo:
Believe it or not the ReRegisterForFinalize behavior is from XNA - in short, when a sound is playing XNA allows the application to drop the instance, which _normally_ causes it to get GC'd, but XNA re-registers as long as the sound is still playing so that it can finish the full effect. Without this, sounds can suddenly cut out and the behavior is random/inconsistent.

7aGiven:
> Believe it or not the ReRegisterForFinalize behavior is from XNA - in short, when a sound is playing XNA allows the application to drop the instance, which _normally_ causes it to get GC'd, but XNA re-registers as long as the sound is still playing so that it can finish the full effect. Without this, sounds can suddenly cut out and the behavior is random/inconsistent.

Thank you for explain. I only reserve the change for DynamicSoundEffectInstance call twice Dispose now.

I will look at the issue and submit another PR.

flibitijibibo:
For the last Dispose, should we change Dispose to not call base.Disposing unless !IsDisposed instead?

7aGiven:
> For the last Dispose, should we change Dispose to not call base.Disposing unless !IsDisposed instead?

`SoundEffectInstance.Dispose(bool disposing)` have checked `!IsDisposed`. I think it's OK for either.

flibitijibibo:
Let's keep the destructor then - I'm usually pretty paranoid about native resource leaks since those are usually what cause console failures the most (aside from excessive disk access). As long as the modification to Dispose works for your game then we should be okay.

7aGiven:
OK, it seem just call Dispose(bool disposing) twice. No error.

Test case:
```C#
using System;

namespace ConsoleApp5
{
    internal static class Show
    {
        public static B obj;
        [STAThread]
        {
            obj = new B();
            obj = null;
        }
    }
    class A
    {
        ~A()
        {
            Console.WriteLine("~A");
        }
    }
    class B : A
    {
        ~B()
        {
            Console.WriteLine("~B");
        }
    }
}
```
output:
```
~B
~A
```

flibitijibibo:
Ohh, yeah this makes sense - dunno why this never occurred to me.

PR #650, opened 2026-08-01
Release GraphicsDevice when GC. Call Disposing when IsDisposed=true. IsDisposed=true during Dispose.

7aGiven:

### DIFF
--- a/src/Graphics/GraphicsDevice.cs
+++ b/src/Graphics/GraphicsDevice.cs
@@ -502,7 +502,48 @@ namespace Microsoft.Xna.Framework.Graphics
 			if (!IsDisposed)
 			{
 				IsDisposed = true;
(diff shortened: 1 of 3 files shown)


src/Graphics/GraphicsDevice.cs
```diff
@@ -502,57 +502,55 @@ protected virtual void Dispose(bool disposing)
 			if (!IsDisposed)
 			{
 				IsDisposed = true;
-
-				if (disposing)
+				/* Dispose of all remaining graphics resources before
```
flibitijibibo, line 505: This function is pretty sensitive so it might take some time to test this - for now I can merge a version of this that only moves the Disposing event call to after DestroyDevice and keeps everything else.

flibitijibibo (changes requested):

PR #618, opened 2026-07-23
Fix GameServiceContainer Exception message

7aGiven:
- [x] I confirm that I am the author of this code and release it to the FNA project under the Ms-PL license. This contribution does not contain code from other sources, including code generated by a Large Language Model ("AI").

PR #657, opened 2026-08-02
GameServiceContainer.AddService Exception message is wrong in XNA

7aGiven:

Found the issue from TestCase output in #618.


7aGiven:
type.GetType().FullName always System.RuntimeType

Change to type.FullName

PR #658, opened 2026-08-02
Optimize Dictionary.TryGetValue in GameServiceContainer.GetService

7aGiven:

> If the key is not found, then the value parameter gets the appropriate default value for the type TValue; for example, 0 (zero) for integer types, false for Boolean types, and null for reference types.

Reference: https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.trygetvalue?view=netframework-4.0

### DIFF
--- a/src/GameServiceContainer.cs
+++ b/src/GameServiceContainer.cs
@@ -61,12 +61,8 @@ namespace Microsoft.Xna.Framework
 			}
 
 			object service;
-			if (services.TryGetValue(type, out service))
-			{
-				return service;
-			}
-
-			return null;
+			services.TryGetValue(type, out service);
+			return service;
 		}
 
 		public void RemoveService(Type type)


7aGiven:
Test Case:
```C#
using System;
using System.Collections.Generic;

namespace ConsoleApp5
{
    static class Show
    {
        [STAThread]
        static void Main()
        {
            object obj;
            Dictionary<Type, object> dictionary = new Dictionary<Type, object>()
            {
                { typeof(B), "B" }
            };
            dictionary.TryGetValue(typeof(A), out obj);
            Console.WriteLine(obj ?? "null");
            dictionary.TryGetValue(typeof(B), out obj);
            Console.WriteLine(obj ?? "null");
        }
    }
    class A { }
    class B { }
}
```
output:
```
null
B
```

PR #603, opened 2026-07-18
Don't throw ArrayIndexOutOfBounds in EffectTechniqueCollection

7aGiven:
- [x] I confirm that I am the author of this code and release it to the FNA project under the Ms-PL license. This contribution does not contain code from other sources, including code generated by a Large Language Model ("AI").

PR #662, opened 2026-08-06
FIXME: ArrayIndexOutOfBounds? in EffectAnnotationCollection and EffectPassCollection. Optimize Argument Check for this[] in EffectParameterCollection and EffectTechniqueCollection.

7aGiven:

The same as #603 #606


7aGiven:
The test case is similar as two PR. So no need write again.

PR #664, opened 2026-08-07
optimize: Microphone

7aGiven:

Maybe don't know how to write test case.

### DIFF
--- a/src/FNAPlatform/SDL3_FNAPlatform.cs
+++ b/src/FNAPlatform/SDL3_FNAPlatform.cs
@@ -1675,16 +1675,12 @@ namespace Microsoft.Xna.Framework
 		 */
 		private static bool micInit = false;
 
-		// FIXME SDL3: This is really sloppy -flibit
-		private static Dictionary<uint, IntPtr> micStreams;
-
 		public static Microphone[] GetMicrophones()
 		{
 			// Init subsystem if needed
 			if (!micInit)
 			{
 				SDL.SDL_InitSubSystem(SDL.SDL_InitFlags.SDL_INIT_AUDIO);
-				micStreams = new Dictionary<uint, IntPtr>();
 				micInit = true;
 			}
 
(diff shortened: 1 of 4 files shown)


flibitijibibo:
I'd have to dig up some old records but oddly enough I think this was hardcoded by XNA for _some_ reason, but I can't recall why... it may have just been that Xbox 360 and Windows Phone headset recordings were a const rate and that carried over to Windows, but I could be wrong on that.

7aGiven:
I think that sampleRate is 44100 when microphone device support 44100. SampleRate is real sampleRate when microphone not support 44100.

flibitijibibo:
Got it, I bet it's a historical quirk then... I could see it always being 44100 for DirectSound/WinXP while WASAPI forces it to react to the device rate more carefully (or at least, that's how it usually is when I'm doing SDL audio work).

7aGiven:
> FIXME: This is what XNA4 aims for, but it _could_ be lower.
Something worth looking at is falling back to lower sample
rates in powers of two, i.e. 44100, 22050, 11025, etc.
-flibit

How to fallback? I have not find a way to test that fallback.

7aGiven:
Wow. The comment exist when the repository created.

7aGiven:
Maybe this is no need. Just draft and think more.

7aGiven:
After look at it.

When recording microphone by request 44100Hz, Windows resampling source to 44100Hz and pass to program.

So do resampling in AudioStream when actual sampleRate not 44100Hz.

7aGiven:
> FIXME: This is what XNA4 aims for, but it _could_ be lower.
Something worth looking at is falling back to lower sample
rates in powers of two, i.e. 44100, 22050, 11025, etc.
-flibit

I think it never fallback in XNA. Windows always resampling though recording device not support 44100Hz.

7aGiven:
> I think it never fallback in XNA. Windows always resampling though recording device not support 44100Hz.

Sorry for my wrong words.

OK, now I finish the research.

The real logical for microphone sampleRate in XNA on Windows Desktop:
```C
WAVEINCAPSW wic;
waveInGetDevCapsW(deviceID, &wic, sizeof(WAVEINCAPSW));

DWORD sampleRate = 0;
DWORD format = wic.dwFormats;
if (format & WAVE_FORMAT_4M16)
    sampleRate = 44100;
else if (format & WAVE_FORMAT_2M16)
    sampleRate = 22050;
else if (format & WAVE_FORMAT_1M16)
    sampleRate = 11025;
```

7aGiven:
Windows Phone always use sample rate 16000 Hz for Windows Phone?

I'm not sure due to Windows Phone Emulator.

7aGiven:
@flibitijibibo How do you think?

Should it fallback to low sample rate? Or just resampling to 44100.

I've never found a way to test XNA on Console. Should it be consider?

src/FNAPlatform/SDL3_FNAPlatform.cs
```diff
@@ -1725,11 +1725,7 @@ ref want
 				SDL.SDL_AudioDevicePaused(audioDeviceID);
 				result[i + 1] = new Microphone(audioDeviceID, name);
 
-				IntPtr stream;
-				SDL.SDL_AudioSpec have;
-				int filler;
-				SDL.SDL_GetAudioDeviceFormat(audioDeviceID, out have, out filler);
-				stream = SDL.SDL_CreateAudioStream(ref want, ref have);
+				IntPtr stream = SDL.SDL_CreateAudioStream(ref want, ref want);
```
flibitijibibo: Is it possible to shrink this via https://wiki.libsdl.org/SDL3/SDL_OpenAudioDeviceStream ?

7aGiven: I think you are right. So I do it in new commit. Remove the micStream Dictionary.

7aGiven: There maybe some bad for checked overflow. Should change SDL_CreateAudioStream return UIntPtr instead of IntPtr?

flibitijibibo: Hmm, good point, this stepped on a lot more than I expected... will have to think about this but the solution may be to error-check inside SDL2_FNAPlatform before creating Microphone objects, so that we don't have to worry about the pointer signage.

7aGiven: I use unchecked keyword now.

Change return type to UIntPtr is optional.

7aGiven: For example:
```C#
					(IntPtr) unchecked((int) SDL.SDL_OpenAudioDevice(
						name,
						1,
						ref want,
						out have,
						0
					),
					))

return (int) SDL.SDL_GetQueuedAudioSize(unchecked((uint) handle));
```

7aGiven:
Don't use object initialize when new struct and pass it to ref argument. It alloc a tmp variable to init and then assign to real variable.

PR #674, opened 2026-08-12
fix: call Disposing when dependency for AudioEngine WaveBank SoundBank

7aGiven:

Test Case:
```C#
using Microsoft.Xna.Framework.Audio;
using System;

namespace ConsoleApp5
{
    static class Show
    {
        [STAThread]
        static void Main()
        {
            AudioEngine audioEngine = new AudioEngine(@"Music.xgs");
            audioEngine.Disposing += AudioEngine_Disposing;

            WaveBank waveBank = new WaveBank(audioEngine, @"Wave Bank.xwb", 0, 512);
            waveBank.Disposing += WaveBank_Disposing;

            SoundBank soundBank = new SoundBank(audioEngine, @"Sound Bank.xsb");
            soundBank.Disposing += SoundBank_Disposing;

            audioEngine.Update();

            Cue cue = soundBank.GetCue("Music_100");
            cue.Disposing += Cue_Disposing;

            Console.Error.WriteLine("start");
            waveBank.Dispose();
            Console.Error.WriteLine("end");
        }

        private static void AudioEngine_Disposing(object sender, EventArgs e)
        {
            Console.Error.WriteLine("AudioEngine_Disposing");
        }

        private static void SoundBank_Disposing(object sender, EventArgs e)
        {
            Console.Error.WriteLine("SoundBank_Disposing");
        }

        private static void WaveBank_Disposing(object sender, EventArgs e)
        {
            Console.Error.WriteLine("WaveBank_Disposing");
        }

        private static void Cue_Disposing(object sender, EventArgs e)
        {
            Console.Error.WriteLine("Cue_Disposing");
        }
    }
}
```
XNA output:
```
start
Cue_Disposing
WaveBank_Disposing
end
```
FNA output:
```
start
WaveBank_Disposing
end
```

### DIFF
--- a/src/Audio/Cue.cs
+++ b/src/Audio/Cue.cs
@@ -279,6 +279,8 @@ namespace Microsoft.Xna.Framework.Audio
 		internal void OnCueDestroyed()
 		{
 			handle = IntPtr.Zero;
+			selfReference = null;
+
 			Dispose();
 		}
 
@@ -295,8 +297,6 @@ namespace Microsoft.Xna.Framework.Audio
 					IsDisposed = true;
 
 					FAudio.FACTCue_Destroy(handle);
-					handle = IntPtr.Zero;
-					selfReference = null;
 
 					if (disposing && Disposing != null)
 					{
--- a/src/Audio/Cue.cs
+++ b/src/Audio/Cue.cs
@@ -278,9 +278,8 @@ namespace Microsoft.Xna.Framework.Audio
 
 		internal void OnCueDestroyed()
 		{
-			IsDisposed = true;
 			handle = IntPtr.Zero;
-			selfReference = null;
+			Dispose();
 		}
 
 		#endregion
@@ -293,12 +292,11 @@ namespace Microsoft.Xna.Framework.Audio
 			{
 				if (!IsDisposed)
 				{
-					// If this is Disposed, stop leaking memory!
-					if (!bank.engine.IsDisposed)
-					{
-						FAudio.FACTCue_Destroy(handle);
-					}
-					OnCueDestroyed();
+					IsDisposed = true;
+
+					FAudio.FACTCue_Destroy(handle);
+					handle = IntPtr.Zero;
+					selfReference = null;
 
 					if (disposing && Disposing != null)
 					{
--- a/src/Audio/SoundBank.cs
+++ b/src/Audio/SoundBank.cs
@@ -142,12 +142,9 @@ namespace Microsoft.Xna.Framework.Audio
 			{
 				if (!IsDisposed)
 				{
-					// If this is disposed, stop leaking memory!
-					if (!engine.IsDisposed)
-					{
-						FAudio.FACTSoundBank_Destroy(handle);
-					}
-					OnSoundBankDestroyed();
+					IsDisposed = true;
+
+					FAudio.FACTSoundBank_Destroy(handle);
 
 					if (disposing && Disposing != null)
 					{
@@ -273,7 +270,6 @@ namespace Microsoft.Xna.Framework.Audio
 
 		internal void OnSoundBankDestroyed()
 		{
-			IsDisposed = true;
 			handle = IntPtr.Zero;
 			selfReference = null;
 			if (dspSettings.pMatrixCoefficients != IntPtr.Zero)
@@ -281,6 +277,8 @@ namespace Microsoft.Xna.Framework.Audio
 				FNAPlatform.Free(dspSettings.pMatrixCoefficients);
 				dspSettings.pMatrixCoefficients = IntPtr.Zero;
 			}
+
+			Dispose();
 		}
 
 		#endregion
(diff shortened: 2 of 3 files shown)


7aGiven:
There is three way to dispose cue.

1. call Cue.Dispose
2. GC call Cue.Finalize
3. call Dispose for WaveBank or SoundBand or AudioEngine and notify OnCueDestroyed

There is one special recursion: OnCueDestroyed be called by FAudio.FACTCue_Destroy(handle) in Cue.Dispose

Disposing and GC.SuppressFinalize should be call when 1 and 3.

7aGiven:
My implementation for the PR maybe not good.
Any idea?

7aGiven:
My satisfactory code that reimplemention for the PR.

flibitijibibo:
Will take a look soon but a common pain point is that the timing of resource destruction also depends on FACT's disposal timing - when a WaveBank is destroyed it tries to dispose the Waves, which in turn destroys Cues, but _doesn't_ destroy SoundBanks since they can make use of multiple WaveBanks - FAudio/src/FACT.c may be of interest for this fix.

7aGiven:
Remove check IsDisposed in the beginning of Dispose. In the past, I think lock keyword not Reentrant Lock. So I delete it now.

7aGiven:
> Will take a look soon but a common pain point is that the timing of resource destruction also depends on FACT's disposal timing - when a WaveBank is destroyed it tries to dispose the Waves, which in turn destroys Cues, but _doesn't_ destroy SoundBanks since they can make use of multiple WaveBanks - FAudio/src/FACT.c may be of interest for this fix.

Thank you, It notify OnXACTNotification that when call FACTCue_Destroy regardless call from native or C#.

7aGiven:
Fix dependency.

WaveBank.Disposing should be called when AudioEngine.Dispose

SoundBank.Disposing should be called when AudioEngine.Dispose

WaveBank and SoundBank allow override Dispose(bool disposing) to not call Disposing.

7aGiven:
`override Dispose(bool disposing)` in WaveBank and SoundBank should be called when AudioEngine.Dispose

7aGiven:
Why do this?
```C#
// If this is disposed, stop leaking memory!
if (!engine.IsDisposed)
{
	FAudio.FACTSoundBank_Destroy(handle);
}
```
If AudioEngine.IsDisposed == true, it mean AudioEngine.Dispose(bool disposing) have been called. FACTAudioEngine_ShutDown will set IsDisposed=true for all SoundBank and WaveBank and Cue.

Maybe this is no need?

flibitijibibo:
Most of what's in the managed side of the XACT implementation is driven by game data, and games behave in really strange ways, especially when exiting. The major problem has been that the FACT engine is often at odds with the CLR, and the GC in particular, since there's already a resource management system in XACT, so we end up doing a lot of extra work that most likely _isn't_ in XNA, but fixes lots of problems across the XACT game catalog.

The last XACT game I shipped was Cattle Country, but there's tons of them out there - Rogue Legacy is another, and probably the most intensive is Dust An Elysian Tail since it swaps out banks frequently for character dialog.

7aGiven:
Some problem:

Is FACT thread safe?

Is FACT allow different thread ID for create thread and destroy thread?

7aGiven:
I notice that Finalize thread is different from Main thread. So I ask these question.

7aGiven:
I try to transfer XNA to FNA for Dust An Elysian Tail.

But it throw Could not load type 'Microsoft.Xna.Framework.GamerServices.Guide' from assembly 'FNA, Version=26.8.0.0, Culture=neutral, PublicKeyToken=null'.

Have not implement GameService?

flibitijibibo:
It probably wants the NetStub: https://github.com/FNA-XNA/FNA.NetStub

On Steam there's also an fna-win32 branch that may help simplify testing.

7aGiven:
> It probably wants the NetStub: https://github.com/FNA-XNA/FNA.NetStub
> 
> On Steam there's also an fna-win32 branch that may help simplify testing.

Thank you, I can get the branch now! It can run successfully.

Myself porting throw System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.

It throw from VideoPlayerTheora.Play

7aGiven:
fna-win32 branch is OK. And I run it with the PR successful. But I don't know how to test to mean the PR is OK?

flibitijibibo:
To help with swapping out FNA I updated Dust's linux_SDL3 branch to also include Windows and macOS builds.

flibitijibibo:
One thing I forgot to mention: XACT is thread-safe, but the catch is that the lock is inside the AudioEngine - once the engine is gone, the mutex is gone.

7aGiven:
> One thing I forgot to mention: XACT is thread-safe, but the catch is that the lock is inside the AudioEngine - once the engine is gone, the mutex is gone.

Thank you, there is some sync problem that need to fix for audio in FNA. Regardless FACT whether support thread safe. I will do a new PR after the other PR be merged.

PR #676, opened 2026-08-13
Fix Icon Load

7aGiven:

### DIFF
--- a/src/FNAPlatform/SDL3_FNAPlatform.cs
+++ b/src/FNAPlatform/SDL3_FNAPlatform.cs
@@ -10,14 +10,15 @@
 #region Using Statements
 using System;
 using System.IO;
+using System.Reflection;
 using System.Text;
 using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


7aGiven:
XNA icon is load from resouce icon from exe.

7aGiven:
The issue occur from Dust An Elysian Tail.

flibitijibibo:
Odd, I thought SDL handled this for us... will have to double check.

7aGiven:
You are right. SDL3 have done it. Is it be done on other platform? I will check.

flibitijibibo:
It doesn't do the same for others, but other platforms also have different icon requirements anyway - macOS for example uses the icns resource instead, and Linux needs a standalone file for xdg desktop shortcuts. Additionally, consoles and other AOT platforms won't have this resource (and likely not a window to have an icon for), so that's why we punt this to SDL rather than decode it ourselves.

PR #680, opened 2026-08-15
optimize: Intersect and Union for Rectangle

7aGiven:

### DIFF
--- a/src/Rectangle.cs
+++ b/src/Rectangle.cs
@@ -397,41 +397,31 @@ namespace Microsoft.Xna.Framework
 			ref Rectangle value2,
 			out Rectangle result
 		) {
-			if (value1.Intersects(value2))
+			int right1 = value1.X + value1.Width;
+			int right2 = value2.X + value2.Width;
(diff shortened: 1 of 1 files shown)


7aGiven:
why operator+ operator- operator* operator/ in Rectangle? it not define in XNA.

src/Rectangle.cs
```diff
@@ -397,41 +397,31 @@ public static void Intersect(
 			ref Rectangle value2,
 			out Rectangle result
 		) {
-			if (value1.Intersects(value2))
+			int right1 = value1.X + value1.Width;
```
flibitijibibo: This took me a second to parse but I understand now - it might make sense to rename/comment in a way that this is clearly an inlining of the Right/Bottom properties. For reference, I did a similar thing in SpriteBatch.PrepRenderState.

flibitijibibo (commented):
Functionally okay, but might need some clarification in the code itself

src/Rectangle.cs
7aGiven: Updated comment

PR #689, opened 2026-08-16
accuracy: Intersects BoundingSphere Ray

7aGiven:

Test Case:
```C#
using System;

namespace ConsoleApp5
{
    static class Show
    {
        [STAThread]
        static void Main()
        {
            BoundingSphere sphere = new BoundingSphere(new Vector3(0f, 0f, 0f), 1f);
            Ray ray = new Ray(new Vector3(1f, 0f, 0f), new Vector3(1f, 0f, 0f));
            float? f = sphere.Intersects(ray);
            Console.WriteLine(f == null ? "null" : f.ToString());
        }
    }
}
```
XNA output:
```
0
```
FNA output:
```
null
```

### DIFF
--- a/src/Ray.cs
+++ b/src/Ray.cs
@@ -244,7 +244,7 @@ namespace Microsoft.Xna.Framework
 			 * the radius of the sphere, it means we've intersected. Checking the
 			 * LengthSquared is faster.
 			 */
-			if (differenceLengthSquared < sphereRadiusSquared)
+			if (differenceLengthSquared <= sphereRadiusSquared)
 			{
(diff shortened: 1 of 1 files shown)


flibitijibibo (commented):
Functionally makes sense, do we need to update the comment above?

7aGiven:
Change comment to less than or equal? I think it's OK.

flibitijibibo:
Yeah, just to be consistent we should change "less than" to "less than or equal" since that appears to have a definite impact on the result

7aGiven:
Updated comment

PR #699, opened 2026-08-19
optimize: reduce Preallocate for FNA3D.cs in .NET and .NET Core 2.1 above

7aGiven:


src/Graphics/FNA3D.cs
```diff
@@ -980,7 +984,7 @@ public static IntPtr ReadImageStream(
 			return pixels;
 		}
 
-		#endregion
+#endregion
```
flibitijibibo: Oh autotabbing, you silly goose

```suggestion
		#endregion
```

flibitijibibo:
Makes sense to me, will check with more experienced .NET Core users in case there's some wacky catch that I'm not aware of

7aGiven:
Thank you, I applied your suggestion.

kg (approved):

PR #701, opened 2026-08-19
optimize: remove redundant encoding for RendererDetail

7aGiven:

### DIFF
--- a/src/Audio/AudioEngine.cs
+++ b/src/Audio/AudioEngine.cs
@@ -134,7 +134,6 @@ namespace Microsoft.Xna.Framework.Audio
 				throw new NoAudioHardwareException();
 			}
 			rendererDetails = new RendererDetail[rendererCount];
-			byte[] converted = new byte[0xFF * sizeof(short)];
 			for (ushort i = 0; i < rendererCount; i += 1)
 			{
 				FAudio.FACTRendererDetails details;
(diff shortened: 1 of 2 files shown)


flibitijibibo:
I think this will break on non-Windows since FAudio unconditionally uses char16_t for these strings, rather than wchar_t.

7aGiven:
> I think this will break on non-Windows since FAudio unconditionally uses char16_t for these strings, rather than wchar_t.

C# System.Char is char16_t in all platform.

FAudio use char16_t instead of wchar_t is right.

7aGiven:
So it's OK for encoding problem.

flibitijibibo:
Sadly on POSIX targets (including some consoles) it's 32-bit, hence the unusual string copies - you should be able to see this on Linux with sizeof, if not CoreCLR then definitely with Mono.

7aGiven:
I'm not sure, but I test it on Ubuntu. It is 16 bit.
图片

7aGiven:
> The [Char](https://learn.microsoft.com/en-us/dotnet/api/system.char?view=net-10.0) structure represents Unicode code points by using UTF-16 encoding. The value of a [Char](https://learn.microsoft.com/en-us/dotnet/api/system.char?view=net-10.0) object is its 16-bit numeric (ordinal) value.
>
>If you aren't familiar with Unicode, scalar values, code points, surrogate pairs, UTF-16, and the [Rune](https://learn.microsoft.com/en-us/dotnet/api/system.text.rune?view=net-10.0) type, see [Introduction to character encoding in .NET](https://learn.microsoft.com/en-us/dotnet/standard/base-types/character-encoding-introduction).

>When a managed [Char](https://learn.microsoft.com/en-us/dotnet/api/system.char?view=net-10.0) type, which is represented as a Unicode UTF-16 encoded code unit, is passed to unmanaged code, the interop marshaller converts the character set to ANSI by default. You can apply the [DllImportAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.dllimportattribute?view=net-10.0) attribute to platform invoke declarations and the [StructLayoutAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.structlayoutattribute?view=net-10.0) attribute to a COM interop declaration to control which character set a marshaled [Char](https://learn.microsoft.com/en-us/dotnet/api/system.char?view=net-10.0) type uses.

reference: https://learn.microsoft.com/en-us/dotnet/api/system.char?view=net-10.0

flibitijibibo (approved):
Was thinking of a different kind of marshaling, checked and this should be fine

PR #711, opened 2026-08-21
Fix Exception for Pan and Apply3D in SoundEffectInstance

7aGiven:

In XNA, Pan and Apply3D allow set with any order before starting sound.

After starting sound. Pan only allow be set when is3D==false, Apply3D only allow be set when is3D==true


7aGiven:
link #710

7aGiven:
Pan and Apply3D are Mutual exclusion.

PR #718, opened 2026-08-22
Implement multi-listener for Apply3D

7aGiven:

### DIFF
--- a/src/Audio/DynamicSoundEffectInstance.cs
+++ b/src/Audio/DynamicSoundEffectInstance.cs
@@ -58,6 +58,18 @@ namespace Microsoft.Xna.Framework.Audio
 
 		#endregion
 
+		#region
+
+		internal override uint Channels
+		{
+			get
+			{
+				return format.nChannels;
+			}
+		}
+
+		#endregion
+
 		#region Internal Variables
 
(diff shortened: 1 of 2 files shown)


flibitijibibo:
My guess is that the F3DAudioCalculate bits are going to change a bit for the multi-listener scenario, so let's just have two separate implementations in the initial draft of multi-listener to avoid that array allocation.

7aGiven:
> My guess is that the F3DAudioCalculate bits are going to change a bit for the multi-listener scenario, so let's just have two separate implementations in the initial draft of multi-listener to avoid that array allocation.

I will implement multi-listener in C# today. If it not work when F3DAudioCalculate not change, Furthermore.

flibitijibibo:
Sounds good - just in case it comes up, our X3DAudio is _mostly_ guesswork, since MSDN doesn't really describe how it calculates things like the output matrix:

https://github.com/FNA-XNA/FAudio/blob/master/src/F3DAudio.c

7aGiven:
Is any way to test the PR?

flibitijibibo:
I don't have a way to test it here - I think this was only supported on 360. Murder Miners is the only game I can think of that _might_ have remnants of it as part of split-screen support but that's a wild guess on my part.

PR #727, opened 2026-08-29
Exception Accuracy for SpriteBatch and SpriteFont. Avoid Array Copy in SpriteFont.ctor

7aGiven:

### DIFF
--- a/src/Graphics/SpriteFont.cs
+++ b/src/Graphics/SpriteFont.cs
@@ -95,7 +95,7 @@ namespace Microsoft.Xna.Framework.Graphics
 
 		/* This is not a part of the spec as far as we know, but we
 		 * added this because it's WAY faster than going to characterMap
-		 * and calling IndexOf on each character.
+		 * and calling BinarySearch for characterMap.
 		 */
 		internal Dictionary<char, int> characterIndexMap;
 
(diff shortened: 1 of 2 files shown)


flibitijibibo:
Similar to #604 I'd like to test this with some long text, like Rogue Legacy's credits screen. I also wonder how much this improves CJK rendering; reminding myself to look at Bastion and Charlie Murder/Dishwasher for this one.

7aGiven:
I think BinarySearch maybe not more speed up than Dictionary (My assume). But reduce memory usage.

The char range is 0-65535, so wrost is compare 16 times for BinarySearch.

7aGiven:
Benchmark show BinarySearch is more slower.

7aGiven:
For avoid slow performance, changed the PR to fix exception and avoid array copy.

flibitijibibo:
Doing a re-review today before tagging tomorrow, should we also replace the BinarySearch here?

https://github.com/FNA-XNA/FNA/commit/249008560f6c1c724874e2d1e248e997106f3523

7aGiven:
> Doing a re-review today before tagging tomorrow, should we also replace the BinarySearch here?
> 
> [2490085](https://github.com/FNA-XNA/FNA/commit/249008560f6c1c724874e2d1e248e997106f3523)

Applied your suggestion

src/Graphics/SpriteFont.cs
```diff
@@ -95,7 +95,7 @@ public float Spacing
 
 		/* This is not a part of the spec as far as we know, but we
 		 * added this because it's WAY faster than going to characterMap
-		 * and calling IndexOf on each character.
+		 * and calling BinarySearch for characterMap.
```
flibitijibibo:
```suggestion
		 * and calling IndexOf on each character. We also tried BinarySearch
		 * with characterMap, but it didn't perform as well.
```

PR #731, opened 2026-09-01
optimize: Remove unused variable initial

7aGiven:


flibitijibibo:
I could be wrong on this, but do we care about the case where one Plane intersects, but a later Plane is Disjoint? This would change behavior to where Intersects would get precedence over Disjoint, not sure if that's intentional.

7aGiven:
> I could be wrong on this, but do we care about the case where one Plane intersects, but a later Plane is Disjoint? This would change behavior to where Intersects would get precedence over Disjoint, not sure if that's intentional.

You are right.

7aGiven:
@flibitijibibo This PR have changed to another PR.

Previous content have been removed by me.

PR #750, opened 2026-09-08
fix: #748

### LINKED ISSUE #748: Rectangle.Intersects is broken
Very simple code:
```c#
using System;

namespace RenderTargetTest
{
	internal class Program
	{
		static void Main(string[] args)
		{
			var r1 = new Rectangle(0, 831, 1990, 514);
			var r2 = new Rectangle(476, 831, 1497, 2520);

			var r = Rectangle.Intersect(r1, r2);

			Console.WriteLine(r);
		}
	}
}
```

with the current FNA it returns `{X:831 Y:831 Width:1142 Height:514}`, which is incorrect.
seems to be broken by this commit: https://github.com/FNA-XNA/FNA/commit/5dd1d513a1c2ea29dbba78cd63d9fba6b8b49e42

If I go before it, then the result is correct: `{X:476 Y:831 Width:1497 Height:514}`

7aGiven:

fix #748

### DIFF
--- a/src/Rectangle.cs
+++ b/src/Rectangle.cs
@@ -403,7 +403,7 @@ namespace Microsoft.Xna.Framework
 			int bottom2 = value2.Y + value2.Height;	// inline value2.Bottom
 			if (value1.X < right2 && value2.X < right1 && value1.Y < bottom2 && value2.Y < bottom1)
 			{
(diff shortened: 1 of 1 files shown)


7aGiven:
link #680

PR #753, opened 2026-09-08
optimize: CopyTo instead of loop for currentlyUpdatingComponents and currentlyDrawingComponents in Game

7aGiven:

Learn from: https://github.com/dotnet/runtime/blob/v8.0.30/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/List.cs#L451

### DIFF
--- a/src/Game.cs
+++ b/src/Game.cs
@@ -185,7 +185,7 @@ namespace Microsoft.Xna.Framework
 		 * -flibit
 		 */
 		private List<IUpdateable> updateableComponents;
-		private List<IUpdateable> currentlyUpdatingComponents;
+		private IUpdateable[] currentlyUpdatingComponents;
 		private List<IDrawable> drawableComponents;
 		private List<IDrawable> currentlyDrawingComponents;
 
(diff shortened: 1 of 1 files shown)


src/Game.cs
```diff
@@ -667,21 +667,29 @@ protected virtual void Draw(GameTime gameTime)
...
-			foreach (IUpdateable updateable in currentlyUpdatingComponents)
+			for (int i = 0; i < capacity; i++)
 			{
+				IUpdateable updateable = currentlyUpdatingComponents[i];
 				if (updateable.Enabled)
 				{
 					updateable.Update(gameTime);
 				}
 			}
-			currentlyUpdatingComponents.Clear();
```
flibitijibibo, line 684: I'd have to dig up the history of this list, but I think this Clear is important since it prevents the objects from having a reference for too long, particularly between a Component being removed from the "real" list and the next Update which clears it. That said, I don't know if List actually does an Array.Clear here, so I could be wrong!

7aGiven: You are right, I add Array.Clear now

7aGiven:
applied your suggestion.

PR #772, opened 2026-09-15
GraphicsDevice.ctor allow PresentationParameters.BackBufferWidth == -1

7aGiven:

The issue found from test case in #768


7aGiven:
link #741

flibitijibibo:
There may need to be a different way to do this; the Viewport property is needed to call FNA3D - do we just need to handle 0-sized viewports instead?

7aGiven:
> There may need to be a different way to do this; the Viewport property is needed to call FNA3D - do we just need to handle 0-sized viewports instead?

OK, let me check. wait.

7aGiven:
Test Case:
```C#
using System;

namespace Show
{
    class Program
    {
        static void Main()
        {
            Game game = new Game();
            PresentationParameters pp = new PresentationParameters() { DeviceWindowHandle = game.Window.Handle, BackBufferWidth = -1, BackBufferHeight = -1 };
            GraphicsDevice gd = new GraphicsDevice(GraphicsAdapter.DefaultAdapter, GraphicsProfile.Reach, pp);

            VertexBuffer vertexBuffer = new VertexBuffer(gd, typeof(VertexPositionColor), 1, BufferUsage.None);
            VertexDeclaration vertexDeclaration = vertexBuffer.VertexDeclaration;
            Console.WriteLine(vertexDeclaration.GraphicsDevice);

            gd.Dispose();
            gd = new GraphicsDevice(GraphicsAdapter.DefaultAdapter, GraphicsProfile.Reach, pp);
            vertexBuffer = new VertexBuffer(gd, vertexDeclaration, 1, BufferUsage.None);
        }
    }
}
```
even if I set BackBufferWidth = -1, it also no error in XNA. So I guess it is for set Viewport later.

flibitijibibo:
apitrace might be able to get a D3D9 capture of XNA to see if it clips the rect or something. Deferring viewport state might be easier said than done, especially if the app never calls it directly after setting a 0-sized backbuffer.

flibitijibibo:
Also: I just made this change to FNA3D, that bug may have been introducing errors depending on the target OS:

https://github.com/FNA-XNA/FNA3D/commit/2c616bf8efd8bfbce288cb1fc7ca3677064c51d9

7aGiven:
OK, I will look at it.

7aGiven:
> There may need to be a different way to do this; the Viewport property is needed to call FNA3D - do we just need to handle 0-sized viewports instead?

You are right, If I set width or height <= 0, after `GraphicsDevice.ctor`, BackBufferWidth and BackBufferHeight and Viewport will to my screen size.

7aGiven:
After looking at it.

XNA transform PresentationParameters to D3DPRESENT_PARAMETERS call IDirect3D9::CreateDevice.
>pPresentationParameters is both an input and an output parameter. Calling this method may change several members including:
>
>    If BackBufferCount, BackBufferWidth, and BackBufferHeight are 0 before the method is called, they will be changed when the method returns.
    If BackBufferFormat equals [D3DFMT_UNKNOWN](https://learn.microsoft.com/en-us/windows/desktop/direct3d9/d3dformat) before the method is called, it will be changed when the method returns.

Then use D3DPRESENT_PARAMETERS to init viewport.
reference: https://learn.microsoft.com/en-us/windows/win32/api/d3d9/nf-d3d9-idirect3d9-createdevice

7aGiven:
So the PR is OK, just let FNA3D_CreateDevice set correct width and height.

7aGiven:
@flibitijibibo I finish the look at for the issue.

flibitijibibo:
Will check everything out on the FNA3D side later today!

7aGiven:
It seem to have not set BackBufferWidth and BackBufferHeight in GraphicsDevice.ctor when them == 0 in PresentationParameters?

7aGiven:
@flibitijibibo Sorry, my words in this PR is wrong, I will do a new PR to fix this, Don't need change FNA3D.

Issue #787, opened 2026-09-18
GraphicsDevice.ctor not set BackBufferWidth and BackBufferHeight when them <= 0


7aGiven:
Test Case:
```
using Microsoft.Xna.Framework;
using Microsoft.Xna.Framework.Graphics;
using System;

namespace Show
{
    class Program
    {
        static void Main()
        {
            Game game = new Game();
            PresentationParameters pp = new PresentationParameters() { DeviceWindowHandle = game.Window.Handle, BackBufferWidth = 0, BackBufferHeight = 0, IsFullScreen = false };
            Console.WriteLine(string.Format("BackBufferWidth={0}, BackBufferHeight={1}", pp.BackBufferWidth, pp.BackBufferHeight));
            GraphicsDevice gd = new GraphicsDevice(GraphicsAdapter.DefaultAdapter, GraphicsProfile.Reach, pp);
            Console.WriteLine(string.Format("BackBufferWidth={0}, BackBufferHeight={1}", pp.BackBufferWidth, pp.BackBufferHeight));
        }
    }
}
```
XNA output:
```
BackBufferWidth=0, BackBufferHeight=0
BackBufferWidth=800, BackBufferHeight=600
```
FNA output:
```
BackBufferWidth=0, BackBufferHeight=0
BackBufferWidth=0, BackBufferHeight=0
```

PR #788, opened 2026-09-18
accuracy: GraphicsDevice.ctor set BackBufferWidth and BackBufferHeight when them <= 0

7aGiven:

link #772
Test Case: #787


flibitijibibo:
Now that this is in place, are we able to call the Viewport property directly again?

7aGiven:
> Now that this is in place, are we able to call the Viewport property directly again?

I think no need validiate Viewport field. If you think it nessessory, you could call the Viewport property directly again. It is fine.

flibitijibibo:
Made that change:

https://github.com/FNA-XNA/FNA/commit/8c3714dcfd09c058443a2bcafd571a0a0e1d2257

I left Reset's alone since the values aren't adjusted like they are in the ctor.

PR #804, opened 2026-09-19
optimize: Array instead of List for ModelMeshCollection

7aGiven:

I don't know why read to `existingInstance`. XNA doesn't do that.

In XNA4, ContentReader read Model with `existingInstance` will throw error.


flibitijibibo:
I wonder if this is a carryover from ancient MonoGame revisions, could be worth doing a git blame on their version of this file to see if there's any historical context.

7aGiven:
> I wonder if this is a carryover from ancient MonoGame revisions, could be worth doing a git blame on their version of this file to see if there's any historical context.

You are right, Monogame support existingInstance and XNA not support. Need compatibility with MonoGame?

7aGiven:
I have known at https://fna-xna.github.io/docs/appendix/Appendix-D%3A-MonoGame/

PR #813, opened 2026-09-20
optimize: Use SharedBuffer with TextureReader and BufferReader and FNA3D.Read and FNA3D.Write and ContentManager.ReadAsset and decompressXNB

7aGiven:

I found XNA do this. I think it is better performance. So I implement a `SharedBuffer` to got it.

The code is reference SharedArrayPool in .NET runtime: https://github.com/dotnet/runtime/blob/v8.0.31/src/libraries/System.Private.CoreLib/src/System/Buffers/SharedArrayPool.cs


flibitijibibo:
Might need thread stress testing but I can see this being a nice way to avoid the GC... if we can have thread-local shared buffers that will likely produce the best result.

7aGiven:
> Might need thread stress testing but I can see this being a nice way to avoid the GC... if we can have thread-local shared buffers that will likely produce the best result.

I use [ThreadStatic] to use ThreadLocal.

7aGiven:
Change class to SharedArray<T> let `currentlyUpdatingComponents` and `currentlyDrawingComponents` to use it.

7aGiven:
link #753

flibitijibibo:
One test case to consider is recursive borrowing - I've always felt bad about this allocation:

https://github.com/FNA-XNA/FNA/blob/master/src/Content/ContentManager.cs#L493-L513

It was a _huge_ performance improvement on Switch 1 though, but maybe this can be improved upon without hammering the GC so much.

7aGiven:
> One test case to consider is recursive borrowing - I've always felt bad about this allocation:
> 
> https://github.com/FNA-XNA/FNA/blob/master/src/Content/ContentManager.cs#L493-L513
> 
> It was a _huge_ performance improvement on Switch 1 though, but maybe this can be improved upon without hammering the GC so much.

You mean decompress xnb file? I know it is streaming decompress in XNA. It could be implements with streaming.

Sure, I think rent a array to decompress is good too.

For recursive borrowing, I only use SharedBuffer with imposible recursive borrowing.

If recursive borrowing possible occur, I won't use SharedBuffer.

7aGiven:
My idea is that:

malloc decompressedSize buffer, and read compressed data to end of buffer.

Then decompress to start of buffer.

7aGiven:
Via XNA, could learn that compressedBuffer and decompressedBuffer is 64KB. Maybe could implements it with in-place decompress? Malloc a buffer that decompressedSize + 64KB? I am not sure.

7aGiven:
Or if 64KB is too large, just malloc a buffer with decompressedSize + compressedSize?

7aGiven:
Need free SharedBuffer?

.NET 8 set a finalizer to check if a buffer not rent for 30s, just free it.

7aGiven:
> Or if 64KB is too large, just malloc a buffer with decompressedSize + compressedSize?

Update the commit implements this.

PR #817, opened 2026-09-23
Remove pChannelAzimuths set

7aGiven:

>Table of channel positions, expressed as an azimuth in radians along the channel radius with respect to the front orientation vector in the plane orthogonal to the top orientation vector. An azimuth of X3DAUDIO_2PI specifies a channel is a low-frequency effects (LFE) channel. LFE channels are positioned at the emitter base and are calculated with respect to pLFECurve only, never pVolumeCurve. pChannelAzimuths must have at least ChannelCount elements, but can be NULL if ChannelCount = 1. The table values must be within 0.0f to X3DAUDIO_2PI. pChannelAzimuths is used with multi-channel emitters for matrix calculations.

reference: https://learn.microsoft.com/en-us/windows/win32/api/x3daudio/ns-x3daudio-x3daudio_emitter

pChannelAzimuths can be NULL if ChannelCount = 1.

if ChannelCount > 1, it will be set by https://github.com/FNA-XNA/FAudio/blob/master/src/FACT3D.c#L89

I have not look at SoundEffectInstance.Apply3D, just wait me to look at it.


7aGiven:
Need consider user reflection to modify emitterData?

PR #818, opened 2026-09-23
optimize: Avoid read from Memory for struct

7aGiven:


flibitijibibo:
This might actually be for @bartwe - the original defensive copy fix was from him, so he's probably the best person to measure this change

bartwe:
This all really depends on what version of dotnet/framework you're targetting.

7aGiven:
Thank you, but Benchmark show this PR is more slow. So I need more thinking.

7aGiven:
> This all really depends on what version of dotnet/framework you're targetting.

@bartwe I look at JIT gen asm for readonly struct field and non-readonly struct field in X86 and X64 in .NET Framework 4.8 and .NET 8 . Have not found anything that non-readonly speed up that readonly. What is issue that you found?

7aGiven:
Only static readonly struct field or directly new in property get could do constant fold, So I think it is better.

But I have not known that why set non-readonly field in that commit? https://github.com/FNA-XNA/FNA/commit/b51a5be6e1cd1a4c8c4b963dabe8e6cff8f4dcf2

flibitijibibo:
The best way to test optimizations will likely be to measure a game's frame time before/after. I might start enforcing this more aggressively for optimizations going forward, since reviewing optimizations shouldn't be done in a vacuum like this.

bartwe:
There are a few things getting mixed here.
1. they are fields to use the trivial property getting pidgeonholed optimizations.
2. they are not readonly to avoid defensive copies, which is needed to avoid full struct copy to stack for accessing a trivial field or calling a method on the struct.

And yes, these optimizations are contextual to the runtime dotnet vs framework vs mono, platform and the program itself.

I agree that your form is far more conductive to constant folding, however it is important to remember this is c#, not c++ . Only quite recent dotnet tries to do such optimizations.

There is also a matter of priority/taste here, in that i also want to ensure the game is realtime performant in debug, something that is perhaps uncommon in the c++ work.

7aGiven:
> The best way to test optimizations will likely be to measure a game's frame time before/after. I might start enforcing this more aggressively for optimizations going forward, since reviewing optimizations shouldn't be done in a vacuum like this.

I have not know that how to ensure warm JIT to test it? Real game how to test it?

7aGiven:
> There are a few things getting mixed here.
> 
>     1. they are fields to use the trivial property getting pidgeonholed optimizations.
> 
>     2. they are not readonly to avoid defensive copies, which is needed to avoid full struct copy to stack for accessing a trivial field or calling a method on the struct.

Thank you. According to your words, I learn that readonly method in C# 8 feature could resolve it. And new commit try to use readonly method to resovle it.

7aGiven:
Use readonly method is better when user set struct in readonly field

7aGiven:
Benchmark to show defensive copy issue:
```C#
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Running;
using System;

public struct Vector4
{
    public static Vector4 One1
    {
        get
        {
            return one;
        }
    }
    public static Vector4 One2
    {
        get
        {
            return oneReadonly;
        }
    }
    public static Vector4 One3
    {
        get
        {
            return new Vector4(1f);
        }
    }

    public float X;
    public float Y;
    public float Z;
    public float W;

    internal static Vector4 one = new Vector4(1f);
    internal static readonly Vector4 oneReadonly = new Vector4(1f);
    private static float random = (float) new Random().NextDouble();

    public Vector4(float value)
    {
        this.X = value;
        this.Y = value;
        this.Z = value;
        this.W = value;
    }

    public float Random()
    {
        return X * random;
    }
}

[DisassemblyDiagnoser]
public class MyBenchmark
{
    [Benchmark]
    public void Vector4One1()
    {
        Vector4.one.Random();
    }
    [Benchmark]
    public void Vector4One2()
    {
        Vector4.oneReadonly.Random();
    }
    [Benchmark]
    public void Vector4One3()
    {
        Vector4.One3.Random();
    }
}

static class Program
{
    [STAThread]
    static void Main()
    {
        BenchmarkRunner.Run<MyBenchmark>();
    }
}
```
## .NET Framework 4.8.1 (4.8.9339.0), X86 LegacyJIT (Job: DefaultJob)

```assembly
; MyBenchmark.Vector4One1()
       mov       ecx,ds:[43C678C]
       add       ecx,4
       call      dword ptr ds:[5B2059C]; Vector4.Random()
       fstp      st(0)
       ret
; Total bytes of code 18
```
```assembly
; Vector4.Random()
       fld       dword ptr [ecx]
       fmul      dword ptr ds:[59552AC]
       ret
; Total bytes of code 9
```

## .NET Framework 4.8.1 (4.8.9339.0), X86 LegacyJIT (Job: DefaultJob)

```assembly
; MyBenchmark.Vector4One2()
       push      edi
       push      esi
       sub       esp,10
       xor       eax,eax
       mov       [esp],eax
       mov       [esp+4],eax
       mov       [esp+8],eax
       mov       [esp+0C],eax
       lea       edi,[esp]
       mov       esi,ds:[3B06790]
       add       esi,4
       movq      xmm0,qword ptr [esi]
       movq      qword ptr [edi],xmm0
       movq      xmm0,qword ptr [esi+8]
       movq      qword ptr [edi+8],xmm0
       lea       ecx,[esp]
       call      dword ptr ds:[4F4059C]; Vector4.Random()
       fstp      st(0)
       add       esp,10
       pop       esi
       pop       edi
       ret
; Total bytes of code 69
```
```assembly
; Vector4.Random()
       fld       dword ptr [ecx]
       fmul      dword ptr ds:[2A652AC]
       ret
; Total bytes of code 9
```

## .NET Framework 4.8.1 (4.8.9339.0), X86 LegacyJIT (Job: DefaultJob)

```assembly
; MyBenchmark.Vector4One3()
       sub       esp,10
       xor       eax,eax
       mov       [esp],eax
       mov       [esp+4],eax
       mov       [esp+8],eax
       mov       [esp+0C],eax
       fld1
       fld1
       fld1
       fld1
       lea       eax,[esp]
       fxch      st(3)
       fstp      dword ptr [eax]
       fxch
       fstp      dword ptr [eax+4]
       fstp      dword ptr [eax+8]
       fstp      dword ptr [eax+0C]
       lea       ecx,[esp]
       call      dword ptr ds:[542059C]; Vector4.Random()
       fstp      st(0)
       add       esp,10
       ret
; Total bytes of code 61
```
```assembly
; Vector4.Random()
       fld       dword ptr [ecx]
       fmul      dword ptr ds:[2E452AC]
       ret
; Total bytes of code 9
```

| Method      | Mean      | Error     | StdDev    | Code Size |
|------------ |----------:|----------:|----------:|----------:|
| Vector4One1 | 0.9550 ns | 0.0039 ns | 0.0035 ns |      27 B |
| Vector4One2 | 1.2543 ns | 0.0069 ns | 0.0064 ns |      78 B |
| Vector4One3 | 1.2632 ns | 0.0092 ns | 0.0082 ns |      70 B |

bartwe:
This specific test shows a clear preference for keeping the original code.

